PR #32, opened 2021-09-08
Sw 27/mission structure

jonathanslee4:


engine/mission.py
```diff
@@ -0,0 +1,41 @@
...
+            grid: the Grid which the robot should traverse
+            grid_mode: "borders" if the grid's nodes should only include corner nodes, "full" if all nodes should be used
+            allowed_dist_error: the maximum distance in meters that the robot can be from a node for the robot to have
+                "visited" that node
+        """
+        self.robot = robot
+        self.grid = grid
+        self.all_waypoints = self.grid.get_waypoints(grid_mode)
+        self.waypoints_to_visit = deque(self.all_waypoints)
+        self.allowed_dist_error = allowed_dist_error
```
sazhang02, line 21: Maybe this should be changed to xyz

sazhang02: no this shoudl not be changed bc xyz

PR #49, opened 2022-02-12
Sw 46/localization

juliehelenek:
https://cornell-nexus.atlassian.net/browse/SW-53?atlOrigin=eyJpIjoiN2ZmMDE0NGUyYTM5NGFjOWExZTBmODUyMWQwODBlYmEiLCJwIjoiaiJ9


sazhang02:
"This task will be completed by @Julie and @Yagmur and supervised by @Jonny and @Jess

Please make sure to add comments in your code so Github reviewers + fellow teammates know what you’re testing for and where.

Upon completion of this feature please detail a general summary of your thought process for your implementation + what exactly your feature does in this Jira ticket here:

Current blocks: for the predict step, we need a control variable which is based on the motor’s movement. Currently we do not have these values, so we are not able to predict. (Current implementation is just updating)

Ready for a PR? Link this Jira ticket in the description or comments of the PR & send a message in slack.

**Note: Make sure to write tests (or some way to validate this feature of working if tests are not possible for EKF) for your code as well (see + click on child issue linked below). You can write these tests on the same branch (or separate if you prefer). Also, I added more child issues than necessary to assign more members.

PR Request Updates 
Right now the localization_visualizer helps us visualize our raw data. So, for example, if we have some GPS data with a bunch of latitude and longitude coordinates, using our visualizer, we can plot it onto the graph and see how accurate our data truly is. When running this file, we take in three parameters: 
python -m engine.localization_visualizer live=0 data=2 ekf=1

live: 0 or 1. 0 means we are getting our data from a file. 1 means we are collecting this data live. Right now, have been focusing on using data from files, so test cases in the future can revolve around live data.

data: 0, 1, or 2. 0 means we are just using imu data, 1 means we are using just gps data, and 2 means we are using both gps and imu data. We will probably be using “2” most frequently. 

ekf: 0 or 1. 0 means we are not using the ekf to make our data smoother, 1 means we are. This is the area in which the pull request is focused.

Our task was to implement the ekf into this visualizer. So, to do this, we would take the raw gps and imu data, and feed it into the ekf and get an output. The big picture of what the ekf does is asking “should we trust our controls (i.e motor) data more or should we trust our sensor (i.e gps/imu) data more?” The ekf does a fancy prediction/linear combination of this values to get an output which is deemed ‘the most accurate’ using all of the data we have at this point. 

The ekf can be broken down into 2 steps, the predict step and the update step. In the predict step, we look at the actual robot itself (i.e it’s motors) and say “Okay, we used this much current to power the motors, which should move the wheels X distance” Right now, we don’t have these values. What this means in terms of our code is mu_bar (The prediction value of mu) is set equal to mu (our current value of mu). Mu is a 3x1 array that holds the robot’s x position, y position, and heading. We did a similar thing for the sigma_bar variable. Sigma is a 3x3 array which is basically ‘the confidence that we have in the system.'  

The update step is where we consider the sensor data and the predictions that we made in the previous step and use these values to make a ‘best guess.' The mathematics behind this was done by Jess and Jonny, but the only change made was to the size of some noise matrices. (This is used for calculations in the update step)

So, what we did in this request was create an if statement saying “if the ekf = 1, we want to do these additional steps” which included initializing the EKF itself. (The initial value for the mu is meters from the starting position, so we used the first gps coordinate and the edge of our grid to determine this) We also used the imu data to determine the starting heading of the ekf. This can be seen in lines 112-124(approx.)

The next step was to use our animation tool to see how the ekf was actually changing our data. So first (ideally) we get a prediction for mu and sigma (these are mu_bar and sigma_bar) Then we get the gps data and imu data for that current time step. We pass in all of these values into the update step, which will update the mu and sigma attributes in the ekf itself (so to plot the new mu, we can access it by doing something like [ekf.mu](http://ekf.mu/)[0][0]) 

We did testing with 2 GPS files. The first is one where the GPS data was taken while walking around the engineering quad. When the ekf was applied to this data, it started off well, but then got stuck at a certain spot. The second GPS data file was taken while standing still in an engineering building. Without the ekf, this data was very jumpy, but when the ekf was used, the data was pretty stationary which is a huge win!"



- Julie from Jira ticket summary :)

PR #55, opened 2022-03-22
added documentation, fixed control mode name mismatch so that test ca…

jessicayuann:
Fixed documentation on the grid file regarding the type of the nodes attribute. Ended up running the grid test file to make sure everything was running and ended up finding an uncaught problem where all test cases were failing due to a mismatch between the control modes listed in grid and those defined in engine.


sazhang02 (approved):
Nice!

PR #61, opened 2022-04-10
SW-51/transmit robot commands addressing TODOs

sazhang02:
made a draft pr so I could comment on the changed files


gui/gui.py
```diff
@@ -33,6 +33,7 @@
 matplotlib.use('TkAgg')
 
 def get_control_mode(window):
+    # TODO: Is this still how we are retrieving control_mode?
```
sazhang02, line 36: Nope you're right. We can prob rename this function to get_sim_control_mode since this is only used in the simulated no threads version OR we can add a flag isSim which will chose whether it does the current code vs read robot_data. This would be done on a diff branch tho

engine/manual.py
```diff
@@ -1,5 +1,53 @@
 # TODO: documentation
 class Manual():
+<<<<<<< HEAD
+    # TODO: move functions from gui and robot here for proper categorization
```
sazhang02, line 4: move or add? if move, which funcs?

engine/manual.py
```diff
@@ -1,5 +1,53 @@
...
+        # will not have access to.
+        elif control_mode == 'Autonomous':
+            self.control_mode = ControlMode()
+
+    def execute_manual(self):
+        robot_data_file = open((self.get_path('csv')[-1] + '/robot_command.csv'), "r+")  # open csv file of robot data
+        # instead of rereading serialized data (which I don't know works), save the data in a field of Manual and pull from there.
+        # byte_data = ser.readline()
+        # data = byte_data.decode()
+        datas = robot_data_file.readlines()
```
sazhang02, line 30: transition to using robot_data aka gui database in gui.py now that we have that?

engine/manual.py
```diff
@@ -1,5 +1,53 @@
 # TODO: documentation
 class Manual():
+<<<<<<< HEAD
+    # TODO: move functions from gui and robot here for proper categorization
+
+    def __init__(self, control_mode):
+        self.control_mode = ControlMode(control_mode)
+
+    # commented part of the code is reading serially. not commented part is testing via csv files
+    def get_control_mode(self):
```
sazhang02, line 10: transition to robot_data

gui/gui.py
```diff
@@ -48,6 +49,41 @@ def get_control_mode(window):
...
+            window['-UP_KEY-'].update(button_color=('black', 'white'))
+            print('forward')
+        elif event == 'd' or event == 'Right':
+            window['-RIGHT_KEY-'].update(button_color=('black', 'white'))
+            print('right')
+        elif event == 's' or event == 'Down':
+            window['-DOWN_KEY-'].update(button_color=('black', 'white'))
+            print('backward')
+        else:
+            # TODO: If option 2 (send command serially with identifier), then have to send unique command to tell the robot to do nothing (doesn't continue moving after key not pressed) - PID stop or just let no power?
```
sazhang02, line 80: good point, we might want to add functions to motorcontroller.py for this

gui/gui.py
```diff
@@ -188,23 +228,36 @@ def setup_gui():
                 [sg.Multiline(current_output, key = "-OUTPUT-", size=(40,8), disabled=True, font=('Courier New', 20))],
                 [sg.Text("Current Coordinates: ______")],
                 [sg.Text("Current Phase: ______", key = "-PHASE-")],
-                [sg.Button('Autonomous', key = "-CONTROL_MODE_BUTTON-"), sg.Button('Track Location'), sg.Button('Traversal Phase'), sg.Button('Simulation')],
+                [sg.Button('CONTROL_MODE', key = "-CONTROL_MODE_BUTTON-"), sg.Button('Track Location'), sg.Button('Traversal Phase'), sg.Button('Simulation')],
```
sazhang02, line 231: Idea but we could make the switch to different control modes a GUI command line command rather than a button. Then if they switch to manual mode we have the buttons appear on screen as mentioned in next todo.

gui/gui.py
```diff
@@ -266,6 +319,12 @@ def run_simulation(name):
 
     window = setup_gui()
     current_row = 0
+    curr_control_mode = "AUTONOMOUS"
```
sazhang02, line 322: why do we need this variable?

gui/transmit_output.py
```diff
@@ -0,0 +1,37 @@
+import os
+import sys
+
+import serial
+
+# ser = serial.Serial("/dev/cu.usbserial-017543DC", 57600)
+# TODO: move functions here from robot and gui for proper categorization. Step 2 would be making sure data is not corrupt
```
sazhang02, line 7: did u mean gui to robot? Seems like this file is for functions that will sned commands from GUI to robot.

gui/gui.py
```diff
@@ -293,7 +352,27 @@ def run_simulation(name):
         if event == 'Simulation':
             simulation_thread = threading.Thread(target=run_simulation, args=(1,), daemon=True)
             simulation_thread.start()
-
+        update_robot_data(window)
+        if event == '-CONTROL_MODE_BUTTON-':
+            robot_data_file = (get_path('csv')[-1] + '/robot_data.csv')
```
sazhang02, line 357: Switch to using robot_data?

engine/robot.py
```diff
@@ -268,7 +268,7 @@ def execute_traversal(self, unvisited_waypoints, allowed_dist_error, base_statio
             self.traverse_roomba(base_station_loc, time_limit, roomba_radius)
         else:
             self.traverse_standard(unvisited_waypoints, allowed_dist_error, database)
-
+        # TODO: read data and if control_mode given is Manual, interrupt other threads. Gui should be sending command serially here? Or should there be some gate which allows commands to be sent serially?
```
sazhang02, line 271: interrupt mission and call execute manual instead?

sazhang02 (commented):
Added comments to discuss

PR #64, opened 2022-04-16
Sw 112/ekf heading

juliehelenek:
https://cornell-nexus.atlassian.net/jira/software/projects/SW/boards/1?assignee=611f1336ee94700071ae746e%2C5f5aa9f10e3968006905cd56&selectedIssue=SW-112

https://cornell-nexus.atlassian.net/jira/software/projects/SW/boards/1?assignee=611f1336ee94700071ae746e%2C5f5aa9f10e3968006905cd56&selectedIssue=SW-109


sazhang02:
"We brainstormed a testing plan as follows:

Do a straight line test with no waypoints, moving at an arbitrary constant velocity

Call the function def motors(self, omega vel) with omega (the heading that we move at) = 0 and vel = c (velocity set to constant value c). Then sleep after 10 seconds

We need to measure the distance the robot travels and the time it takes for the robot to travel this distance 

With these values, we can get the controls for these time steps 

We also need to measure the GPS&IMU data at every loop

We can use phone GPS or other resources to read the start and end position of the robot and compare the numeric accuracy with the EKF 



The testing script calls the motors functions, collects gps data, and collects imu data at every time step." - julie's jira ticket summary

PR #70, opened 2022-04-19
implement live vs sim

paulhsuh:
do not merge (draft pr)


AC408:
## References
PR Number: #70 
Jira Ticket: [SW-103](https://cornell-nexus.atlassian.net/browse/SW-103?atlOrigin=eyJpIjoiZWNlNzRmMzcyNDI1NDgzMmIzZGQzZmMzZjlhM2Y1NzMiLCJwIjoiaiJ9)

## Summary
### TLDR: differentiated between live and sim (motors turn only when live and not sim; sim simulation runs only when sim and not live)

### Description
Added flags for live vs sim and add relevant code depending on whether flag is toggled to live or sim (motors moving, etc) in robot.py. Added a script to test the PID. Tests added includes no PID straight line test (to see if robot is able to move and if the direction is right), PID straight line test (to test PID on the most simple path), and PID grid straight line test (to see if PID can handle a graph as input). Updated parameters for motors (vm_load). Resolved merge conflicts (at least with gui with the control mode buttons). Added additional parameters to robot.py to initialize GPS and IMU for testing without needing to set up everything. Added comments and debug printing. 

## Test Plan
Changes works because running the script with the live toggle on runs the motor and with the live toggle off doesn't run the motor. 

## Other
Co-author: Paul :'(

engine/robot.py
```diff
@@ -202,6 +240,13 @@ def move_to_target_node(self, target, allowed_dist_error, database):
                     str(self.state[0])[1:-1] + ',' + str(self.state[1])[1:-1] + ',' + str(self.state[2])[1:-1] + '\n')
             time.sleep(0.001)
 
+            if not self.is_sim:
+                print("state")
+                print(self.state)
+                self.state = self.update_ekf_step()
+                print("ekf_state")
+                print(self.state)
```
AC408, line 248:
```suggestion
                self.state = self.update_ekf_step()
```

engine/robot.py
```diff
@@ -264,12 +314,30 @@ def print_current_state(self):
...
 
     def execute_setup(self, radio_session, gps, imu, motor_controller):
-        gps_setup = gps.setup() 
+        gps_setup = gps.setup()
         imu_setup = imu.setup()
         radio_session.setup_robot()
-        motor_controller.setup()
+        motor_controller.setup(self.is_sim)
+
+        zone = ENGINEERING_QUAD  # Used for GPS visualization
```
AC408, line 322:
```suggestion
        zone = ENGINEERING_QUAD  # Used for GPS visualization, make it a parameter
```

engine/robot.py
```diff
@@ -132,9 +145,30 @@ def __init__(self, x_pos, y_pos, heading, epsilon, max_v, radius, is_sim=True, p
         with open(cd + '/phases.csv', 'a') as fd:
             fd.write(str(self.phase) + '\n')
 
-    def travel(self, dist, turn_angle):
+    def update_ekf_step(self):
+        zone = ENGINEERING_QUAD  # Used for GPS visualization
```
AC408, line 149:
```suggestion
        zone = ENGINEERING_QUAD  # Used for GPS visualization, make this not hard-coded
```

gui/gui.py
```diff
@@ -48,6 +48,45 @@ def get_control_mode(window):
 
     file.close()
 
+def manual_mode_actions(window, event, curr_control_mode):
```
AC408, line 51: honestly shouldnt include changes in this file in this pr

tests/functionality_tests/pid_tuning.py
```diff
@@ -1,20 +1,31 @@
...
+import time
+
+from electrical.gps import GPS
+from electrical.imu import IMU
+
+import serial
+import busio
+import board
+
+import csv
```
AC408, line 20: not sure where serial, busio, and board is used. I think they were used when using the other motorcontroller class. looks like there's an extra import csv

AC408 (commented):
uhh, this pr doesn't show changes from beginning of branch. and previous pr doesn't follow this pr's format, so there could be some changes that should be in main that escapes this pr

juliehelenek (commented):
good work! just added a few comments for documentation (:

engine/robot.py
```diff
@@ -132,9 +145,30 @@ def __init__(self, x_pos, y_pos, heading, epsilon, max_v, radius, is_sim=True, p
...
+        new_state = np.array([[new_x], [new_y], [new_heading]])
+        return new_state
+
+    def travel(self, velocity, omega):
         # Moves the robot with both linear and angular velocity
-        self.state = np.round(integrate_odom(self.state, dist, turn_angle), 3)
+        self.state = np.round(integrate_odom(
+            self.state, velocity, omega), 3)
         # if it is a simulation,
         if self.is_sim:
```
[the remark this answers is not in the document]
AC408: that's the logic for it right now. We could move `mc.motors(limited_cmd_w[0], limited_cmd_v[0])` from line 228 to this not sim branch

tests/functionality_tests/pid_tuning.py
```diff
@@ -67,4 +83,21 @@ def animate(i):
...
 
 
+def one_node_straight_line():
+    mc = MotorController()
+    gps = None
+    init_gps = None
+    imu = None
+    r2d2 = Robot(0, 0, math.pi / 2, epsilon=.2, max_v=.2, radius=5, init_phase=Phase.TRAVERSE)
+    database = DataBase(r2d2)
+    r2d2.move_to_target_node([0, 1], .03, database, mc, gps, init_gps, imu)
```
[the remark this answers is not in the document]
AC408: Why would it move forever? Wouldn't the robot still keep moving until it reaches that coordinate? But if you're talking about the parameter being too large, yes that's probably the case and either way, it's probably better to pass in a variable/parameter anyways.

tests/functionality_tests/pid_tuning.py
```diff
@@ -67,4 +83,21 @@ def animate(i):
...
+
+
+def no_pid_straight():
```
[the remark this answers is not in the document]
AC408: yes, this should move the car forward at a speed of .2 for 5 seconds.

gui/gui.py
sazhang02: oof sorry github didn't send me an email thought it would've. @AC408 could you add this to a separate PR as you mentioned? It is pretty separate from the changes in this PR and I think it would benefit to test this independently (+ bc potential merge conflicts w/ the rpi to GUI pr). In general tho the code for this GUI file seems good to me!

tests/functionality_tests/pid_tuning.py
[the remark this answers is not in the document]
sazhang02: @juliehelenek

sazhang02:
@yagmurdulger

AC408:
removed gui changes, merged main, and resolved merge conflicts.

tests/functionality_tests/pid_tuning.py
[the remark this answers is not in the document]
juliehelenek: move to target node is going redevelopment, so resolving this

juliehelenek (approved):
thank you for reviewing the comments ~~ we should come back to this once the traversal logic is retested!

electrical/gps.py
```diff
@@ -4,7 +4,7 @@
 import time
 import pynmea2
 import csv
-if False: 
+if True: # import UbloxGPS if this is not a simulation
```
yagmurdulger, line 7: Change this to `if False` because otherwise test cases on our laptop will fail. In the future, we want to avoid hardcoding this (onboarding task).

electrical/imu.py
```diff
@@ -1,5 +1,5 @@
 import time
-if False: 
+if True:
```
yagmurdulger, line 2: Same as above. Change to `if False`.

engine/robot.py
```diff
@@ -152,10 +186,9 @@ def move_forward(self, dist):
             self.truthpose = np.append(
                 self.truthpose, np.transpose(self.state), 0)
 
-    def move_to_target_node(self, target, allowed_dist_error, database):
+    def move_to_target_node(self, target, allowed_dist_error, database, mc):
```
yagmurdulger, line 189: instead of adding a new `mc` parameter here, can we use the motor_controller attribute of the robot class?

yagmurdulger (commented):
I think this looks pretty good! I just have a few minor comments. After you do those, if it passes all the test cases, it's good to merge. Great job :)

yagmurdulger (dismissed):
Commented above

sazhang02:
@AC408 oops can u resolve merge conflicts and then ill force merge this

sazhang02 (approved):

PR #81, opened 2022-09-10
fixed get_path

AC408:
Made it so get_path isn't relative to file the function is called in, but relative to src also updated todos

## References
N/A

### TLDR: 1-3 sentence TLDR here
Added a top-level definition.py file to hold the path to the root directory (src) and used that whenever files want to get a path (relative to src). Also added todos to other tasks in robot.py

### Description
See summary.

## Test Plan
Ran traverse_standard_test under robot_test under compilation_test and path no longer pulled (or tried to pull) from .../compilation_test/robot_test/csv. Also ran python3 -m gui.gui and ran the simulation and data pulled from right file (because gui was also changed to reflect this update).

## Other
N/A


engine/robot.py
```diff
@@ -126,10 +118,9 @@ def __init__(self, x_pos, y_pos, heading, epsilon, max_v, radius, is_sim=True, p
             output_limits=(None, None)
         )
 
-        cwd = os.getcwd()
-        cd = cwd + "/csv"
+        # TODO: wrap in try/except (error when calling execute_setup_test.py)
```
sazhang02, line 121: Let's keep these todo comments in a separate pr next time

sazhang02 (approved):
Great abstraction and job with the PR!!!

sazhang02:
@juliehelenek

tests/compilation_tests/mag_heading_test.py
```diff
@@ -4,16 +4,16 @@
 import numpy as np
 from ahrs.filters import EKF
 import csv
-import os 
+import os
```
juliehelenek, line 7: adding a comment here because i cant figure out how to put it above:

running this test _only works_ when i remove the wmm2020 import; wmm2020 isn't even used in this test so i think that we can just remove it

otherwise, the pathing worked on my computer!

engine/robot.py
juliehelenek: ^! we dont want to forget them so make sure these TODOs are mentioned in some sort of future steps documentation / mentioned in person to members

juliehelenek (commented):
just 1 change! rn running doesnt work for me because of the wmm2020 import

tests/compilation_tests/mag_heading_test.py
sazhang02: good point! lets do this in a separate pr, could you make it @juliehelenek ?

juliehelenek (approved):

PR #86, opened 2022-10-03
Created Odometry Class 

angelaloh09:
## References
- Odometry Calculations in odometry() method found here: https://drive.google.com/drive/u/1/search?q=odometry

## Summary
Created odometry class. The goal is to be able to collect data from the back motor encoders and understand the change in angular displacement and linear displacement of the robot. This data will eventually be fed back into the PID control loop for the motors and used to tune the motors' performances. 

Also updated the motor controller class with some slight changes, since it was a bit outdated on main. Jerry and I made some small changes to the motor controller code when testing the motors with the new controller. 

## Test Plan
TODO @404coderCN


electrical/motor_controller.py
```diff
@@ -144,20 +140,3 @@ def motors(self, omega, vel):
...
-            GPIO.setup([self.enA, self.enB], GPIO.OUT)  # EnA, EnB
-
-            self.p1 = GPIO.PWM(self.enA, 50)
-            self.p2 = GPIO.PWM(self.enB, 50)
-
-            # Initialize PWM duty cycles as 0
-            self.p1.start(0)
-            self.p2.start(0)
-
-            self.motors(0, 0)
```
sazhang02, line 163: Currently `robot.py` and `robot_test.py` use this function. Is there an alternative function we can call instead of this?

sazhang02: update nvm -- julie pointed out the new setup func in the other pr!

sazhang02:
Is this pr ready to be reviewed/merged? Just wondering since the test plan says todo @ jerry

angelaloh09:
PR not ready yet, just because we tested the code -- while we can get encoder data back, the behavior isn't usual. Jerry and I are in the process of debugging whether this is a hardware issue or software issue!

sazhang02:
uhh im not sure what happened oops

PR #87, opened 2022-10-05
Adding Tests to Grid_Test

Ellied5:
### Description
This is a few tests to improve code coverage on grid.py.

## Test Plan
This is consists of testing.


csv/phases.csv
```diff
@@ -0,0 +1,55 @@
+Phase.TRAVERSE
```
sazhang02, line 1: Could you clear csv/phases.csv, csv/datastore.csv, and rpi_to_gui.simulation.csv? A lil context on this, but we have our code write to these csv files when the code is run to pass information between our engine and GUI modules. This happens whenever our code is run, including tests. However, we usually want to keep them empty on our main branch for storage reasons. Just a select all + delete for the files

tests/compilation_tests/mag_heading_test.py
```diff
@@ -1,4 +1,4 @@
-import wmm2020
+#import wmm2020
```
sazhang02, line 1: :D 👍

tests/compilation_tests/grid_test.py
```diff
@@ -98,8 +101,20 @@ def test_node_boundaries(self):
...
+
+
+class TestInside(unittest.TestCase):
+    def test_inside(self):
+        lat_min, lat_max, long_min, long_max = 42.444250, 42.444599, -76.483682, -76.483276
+        g = Grid(lat_min, lat_max, long_min, long_max)
+        self.assertTrue(g.isInsideTriangle(2, 2, 3, 3, 1, 3, 2,
+                        2.5), "this point is inside the triangle")
+        self.assertFalse(g.isInsideTriangle(2, 2, 3, 3, 1, 3,
+                         10, 10), "this point is outside the triangle")
```
sazhang02, line 117: I think it would be good to add this function inside of `TestGrid` since the extent of `TestInside` would be to test this helper function.

tests/compilation_tests/grid_test.py
```diff
@@ -98,8 +101,20 @@ def test_node_boundaries(self):
...
-        self.assertLessEqual(top_right_node.get_m_coords()[0], x_range, "The meters grid shouldn't be larger than the lat bounds")
-        self.assertLessEqual(top_right_node.get_m_coords()[1], y_range, "The meters grid shouldn't be larger than the long bounds")
+        self.assertLessEqual(top_right_node.get_m_coords()[
+                             0], x_range, "The meters grid shouldn't be larger than the lat bounds")
+        self.assertLessEqual(top_right_node.get_m_coords()[
+                             1], y_range, "The meters grid shouldn't be larger than the long bounds")
+
+
+class TestInside(unittest.TestCase):
+    def test_inside(self):
```
sazhang02, line 111: Could we add a case for a point that is on the edge of a triangle? 🤔

sazhang02 (changes requested):
AMAZING job Ellie on your first PR! 🎉 🎉 🎉  Suggested some small changes below and it should be good to merge afterward 🎉 

Some additional PR suggestions for future reference (a bit on the longer side for explanation purposes!):
- Github has a cool feature called Draft PRs (I recently discovered this like last sem), which allows you to make a PR but indicates it's still WIP (while PR means ready to merge to main) and you can turn it into a PR later. Think it would be a neat way to get suggestions from fellow teammates when you're not completely done. I believe when you have the option to make a PR, there's a lil drop-down arrow next to it to make it a draft instead if u need to make on in the future)
- Suggestion for branch name to be a lil more specific just bc `testing` might overlap with someone else's branch name, `test_inside_triangle` would be good
  - Likewise for the PR name something like "Add tests for isInsideTriangle to grid_test" would be good in the long run to help future members search old prs (having those keywords for control f-ing LOL)!
- In this case, although you are adding a test, I would add to the summary of the PR the steps you took to make sure the test was added correctly. For example as you mentioned to me earlier, you ran `python -m tests.compilation_tests.test_runner` and compared the code coverage (+ maybe insert ss of new code coverage :00000). Also could mention the wmm2020 problem you found and point it out as something that needs to be fixed!

tests/compilation_tests/grid_test.py
sazhang02: Also something I just realized 🤔 `isInsideTriangle` is in camel case while the rest of our code is in snake case. As a part of a ~realization~ resulting from this test, I think it would be good to fix this!

csv/phases.csv
AdityaKompella6: Sammy made a great comment.

tests/compilation_tests/grid_test.py
AdityaKompella6: Godd jobb so I think that a test for the circles  that check the extremes also makes sense

tests/compilation_tests/grid_test.py
AdityaKompella6: Yep

AdityaKompella6 (approved):

sazhang02 (approved):
Looks good to me! Just added back the csv files (since we just needed to empty its contents). Ready for you to merge o7

PR #88, opened 2022-10-05
Adding grid isOnBorder() tests and deleting duplicate method

Aayush-Agnihotri:
### Description

Replace this with description
image

### DIFF
--- a/engine/grid.py
+++ b/engine/grid.py
@@ -161,31 +161,6 @@ class Grid:
 
     # --------------------- METHODS TO FINISH INITIALIZATION OF ACTIVATED GRID -------------- #
 
-    def is_on_border(self, row, col, row_limit, col_limit):
-        """
-        Returns whether a particular activated node is on the border.
-
-        An activated node is on the border if any of the following conditions hold:
-        1. Any of its neighboring nodes are inactive.
-        2. It exists on the very edge of the grid.
-        """
-        min_col = max(0, col-1)
-        min_row = max(0, row-1)
-        max_col = min(col_limit, col+1)
-        max_row = min(row_limit, row+1)
-
-        # If this node is on the very edge of the grid, it is automatically a border node
-        if min_col == 0 or min_row == 0 or max_col == col_limit or max_row == row_limit:
-            return True
-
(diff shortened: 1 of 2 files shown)


engine/grid.py
```diff
@@ -161,31 +161,6 @@ def determine_active_waypoints(self, node):
 
     # --------------------- METHODS TO FINISH INITIALIZATION OF ACTIVATED GRID -------------- #
 
-    def is_on_border(self, row, col, row_limit, col_limit):
```
sazhang02, line 164: Nice! 👍

tests/compilation_tests/grid_test.py
```diff
@@ -4,6 +4,7 @@
 from engine.grid import Grid
 from engine.kinematics import get_vincenty_x, get_vincenty_y
 from engine.mission import ControlMode
+from tests.functionality_tests.test_activation_shapes import is_on_border
```
sazhang02, line 7: Since you removed the duplicate is_on_border, I think we can delete this import statement and use the remaining method from engine.grid (which is already imported)

sazhang02 (changes requested):
FANTASTIC job Aayush on your first PR! 🎉 🎉 🎉  One very small change below and it should be good to merge afterward 🎉 

Some additional PR suggestions for future reference (a bit on the longer side for explanation purposes!):
- Github has a cool feature called Draft PRs (I recently discovered this like last sem), which allows you to make a PR but indicates it's still WIP (while PR means ready to merge to main) and you can turn it into a PR later. Think it would be a neat way to get suggestions from fellow teammates when you're not completely done. I believe when you have the option to make a PR, there's a lil drop-down arrow next to it to make it a draft instead if u need to make on in the future)
- In addition to the nice code cov image, I would add what terminal commands you ran to get that for whoever's reviewing to see so they can reproduce

engine/grid.py
AdityaKompella6: I love this deletion exquisite!

AdityaKompella6 (approved):
Beaueitiful

sazhang02 (approved):
Nice!!! Let's merge this :DDDD

PR #90, opened 2022-10-12
Add tests for remaining database.py functions (phase_as_value + make_packet)

sazhang02:
### Description

In order to increase our unit testing, I added some tests for the remaining functions in `database.py`. 

In `test_phase_as_value`, I created a database instance and tests `phase_as_value` against a list of all phase values, updating the phase stored in the database respectively. Made sure to use our enum `Phase` here in case we modify the enum in the future. 

In `test_make_packet`, I created the expected packet (specified in `packet.py`) for each of the global test databases. Note, the test cases for `test_make_packet` can be expanded in the future once we determine how to extract our packet properties from our main mission (e.g. rot, last_n which isn't a direct property of robot atm). I also came across a bug where acceleration was being converted to a string incorrectly since we were adding rather than appending strings to a list & fixed it.

Code Coverage:
Before: 60%
Screen Shot 2022-10-12 at 7 25 24 PM
After: 100% \o/
Screen Shot 2022-10-12 at 7 25 38 PM

## Test Plan

Ran `coverage run -m tests.compilation_tests.test_runner` to run our compilation tests and make sure the new tests don't fail
`coverage html `,`open htmlcov/index.html ` to see coverage changes.

### DIFF
--- a/engine/database.py
+++ b/engine/database.py
@@ -93,7 +93,7 @@ class DataBase:
         
         acc = []
         for i in self.get_data("acceleration"):
-            acc += str(i)
+            acc.append(str(i))
 
         temp_n_dist = "00.0"
         temp_rot = "00.00"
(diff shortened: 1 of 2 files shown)


sazhang02:
Don't go easy on me guys 😤

jinlin270 (approved):

PR #93, opened 2022-10-15
Integrate motor controller

yagmurdulger:
Prior to this PR, our `move_to_target_node` and `turn_to_target_heading` functions weren't integrated with the electrical `motor_controller` code, and so, all movement was simulated. With these changes:
- Robot class is able to call `spin_motors` function with the desired (clamped) linear and angular velocities as parameters
- The motor controller then converts this information into different voltages that should be applied to the left and right motors for desired movement.
- Additionally, we now initialize `motor_controller` in `Robot` rather than `Mission` which enables the above functionality.

### DIFF
--- a/engine/mission.py
+++ b/engine/mission.py
@@ -64,7 +64,7 @@ class Mission:
         self.inactive_waypoints = self.grid.get_inactive_waypoints_list()
         self.waypoints_to_visit = deque(self.all_waypoints)
         self.allowed_dist_error = allowed_dist_error
-        if self.robot.is_sim:
+        if not self.robot.is_sim:
             self.gps_serial = serial.Serial('/dev/ttyACM0', 19200, timeout=5)
             self.robot_radio_serial = serial.Serial(
                 '/dev/ttyS0', 57600)  # robot radio
--- a/engine/mission.py
+++ b/engine/mission.py
@@ -33,6 +33,7 @@ Commented out for simulation testing purposes
 # import busio
 # import adafruit_lsm9ds1
 
+
 class Mission:
     def __init__(self, robot, base_station, init_control_mode, grid=Grid(42.444250, 42.444599, -76.483682, -76.483276),
                  allowed_dist_error=0.5, allowed_heading_error=0.1, allowed_docking_pos_error=0.1,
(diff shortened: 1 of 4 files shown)


engine/mission.py
```diff
@@ -63,32 +64,34 @@ def __init__(self, robot, base_station, init_control_mode, grid=Grid(42.444250,
         self.inactive_waypoints = self.grid.get_inactive_waypoints_list()
         self.waypoints_to_visit = deque(self.all_waypoints)
         self.allowed_dist_error = allowed_dist_error
-        if not robot.is_sim:
-            self.gps_serial = serial.Serial('/dev/ttyACM0', 19200, timeout=5) 
-            self.robot_radio_serial = serial.Serial('/dev/ttyS0', 57600) #robot radio 
+        if self.robot.is_sim:
```
juliehelenek, line 67: should be if not self.robot.is_sim

engine/mission.py
```diff
@@ -63,32 +64,34 @@ def __init__(self, robot, base_station, init_control_mode, grid=Grid(42.444250,
...
-
     def execute_mission(self, database):
         """
         Activates the main control loop. Depending on the robot's phase, different motion control algorithms are
         activated.
         """
         while self.robot.phase != Phase.COMPLETE:
             if self.robot.phase == Phase.SETUP:
-                self.robot.execute_setup(self.robot_radio_session, self.gps, self.imu, self.motor_controller)
+                self.robot.execute_setup(
```
juliehelenek, line 93: is the spacing weird when u open in vs code?

engine/robot.py
```diff
@@ -102,6 +104,9 @@ def __init__(self, x_pos, y_pos, heading, epsilon, max_v, radius, is_sim=True, p
         self.gyro_rotation = [0, 0, 0]  # TEMPORARY
         self.linear_v = 0
         self.angular_v = 0
+        if not self.is_sim:
```
juliehelenek, line 107: do we still want to initialize one when we don't have a simulation?

juliehelenek: why are we only initializing one if its not a simulation, but then later in robot.py we only use the motorcontroller if it IS a simulation?

angelaloh09 (changes requested):
some small changes to be made!

angelaloh09 (approved):
looks good to me

PR #94, opened 2022-10-16
Refactor activation test

jinlin270:
### Currently, test_activation_shapes.py tests the activation_rectangle, activation_triangle, and activation_circle.  functionalities in grid.py. The test will draw a graph of a grid of nodes with a rectangle/triangle/circle shape in the middle, depending on which shape is input as the activation type. All the nodes inside the rectangle/triangle/circle shape are activated (green), the nodes on the edge of the shape are border_nodes (red), and all the nodes not in the shape are unactivated (blue). There are also purple lines within the activated area representing the traversal path. 
Because all parameters are hard-coded, the graphs produced by test_activation_shapes.py should always look the same. This means test_activation_shapes.py can serve as a reference and basic check for future implementations to make sure the new code doesn't produce unwanted drastic changes in Grid. 

### Description

We refactored test_activation_shapes.py to use functions in grid.py, and we removed all duplicated code. We also fixed a bug in Grid.py to remove the assumption that a Node object's positions are always integers. The code works, but it seems like in grid.py, the rows are corresponding to the x-axis, and the columns are corresponding to the y-axis. This does not make logical sense, so we are thinking about it.

Here are what the graph for each activation_type should look like:
Rectangle:
Screen Shot 2022-10-16 at 5 23 38 PM

Triangle:
Screen Shot 2022-10-16 at 5 24 29 PM

Circle:
Screen Shot 2022-10-16 at 5 24 48 PM

## Other

Edit: I have removed is_raspberrypi.py from this PR and added the 3 graphs to tests/functionality_tests/activate_shape_graphs. I have also deleted lines not in Aditya's original code in the file retired/2022-2023/test_activation_shapes.py.


engine/is_raspberrypi.py
```diff
@@ -0,0 +1,12 @@
...
+"""a function from : https://raspberrypi.stackexchange.com/questions/5100/detect-that-a-python-program-is-running-on-the-pi/74541#74541"""
+def is_raspberrypi():
+    try:
+        with io.open('/sys/firmware/devicetree/base/model', 'r') as m:
+            if 'raspberry pi' in m.read().lower(): return True
+    except Exception: pass
+    return False
+
+
+print(os.uname())
```
sazhang02, line 12: Let's take this file out of this PR and put it on a new branch (since checking whether you're on a raspberry pi is a diff task).

retired/2022-2023/test_activation_shapes.py
```diff
@@ -1,7 +1,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 from enum import Enum
-
+from engine.grid import Grid
```
sazhang02, line 4: I think we can delete this line and 164 since it wasn't used in the original file Aditya made (prob got added when we were testing stuff between this and the test file)

retired/2022-2023/test_activation_shapes.py
```diff
@@ -161,6 +161,7 @@ class WaypointPhase(Enum):
 
 
 def main():
+    grid=Grid(42.444250, 42.444599, -76.483682, -76.483276)
```
sazhang02, line 164: see above

engine/grid.py
```diff
@@ -319,12 +319,13 @@ def find_border_nodes(self):
...
-        candidate_nodes = [node for node in self.border_nodes if node.y == pos[1]+1]
+        # node_info: (node, row, col)
+        candidate_nodes = [node_info for node_info in self.border_nodes if node_info[2] == pos[1]+1]
         if (candidate_nodes == []):
             return None
         else:
-            node = min(candidate_nodes,key=lambda node: node.x)
-            return (node.x,node.y)
+            node_info = min(candidate_nodes,key=lambda node_info: node_info[1])
+            return (node_info[1],node_info[2])
```
sazhang02, line 328: @AdityaKompella6 This (and the respective adjustments in find_border_nodes when making border_list) is where we found the "bug" Lin mentioned in the PR description. Could you double-check this to make sure this makes sense with you?

So, rather than using a nodes x,y values (as it may not be a perfect integer, but rather a float value as we are creating a grid from lat,long values) we changed it to use row, column values as we realized the test_activation_shapes script was set up to initialize nodes x,y's using row, column.

sazhang02: @jinlin270 I think I get the row, col -> x/y-axis relationship in this function (or maybe this is a 3am brain fart)

node_info[1] corresponds to the row this node is located (aka y-value). 
node_info[2] corresponds to the column this node is located (aka x-value).

I think here and in `get_all_lawnmower_waypoints_adjustable` we are traversing **vertically** (rather than horizontally like we discussed in full team).

So, `node_info[2] == pos[1]+1` searches for all the nodes in the column to the right of the current column. Then, `node_info = min(candidate_nodes,key=lambda node_info: node_info[1])` finds the node out of those with the lowest row. So, this new bottom_rightmost_node will be used in 'get_all_lawnmower_waypoints_adjustable', which iterates over this node and all the ones above it in the same column.

@AdityaKompella6 if you could confirm this that would be amazing 🛐

sazhang02 (changes requested):
Looks good to me!!! Super nice job with the PR and description :D Just some small comments from my end (but I also helped write this code so maybe I'm biased LOL).

An idea I just had would also be to make a folder within test/functionality_tests where we put the expected images (the three you showed in the PR), so we can reference them as we run the scripts. Would be cool if you could also add this :3

engine/grid.py
AdityaKompella6: Yep float to int makes sense. Also starting at the bottom right is correct.

AdityaKompella6 (approved):
Yep the two changes you made to find the rightmost border and float coords to integer are right

sazhang02:
YAYY LOOKS GREAT!!!

sazhang02 (approved):
Approving! Once merge conflict fixed u can merge \o/

PR #96, opened 2022-10-22
added obstacle tracking and obstacle following main algorithm

AC408:
## References
https://docs.google.com/document/d/1jpvVf5tuF9DO4oXTGWUN_emfemW88B4P29G1xxR9C0w/edit?usp=sharing

### TLDR: Replace this with 1-3 sentence TLDR
Currently, no obstacle avoidance. We implement tangent bug algorithm. We have a separate thread to detect if there's an obstacle and the main obstacle avoiding algorithm.

### Description
- Obstacle tracking detects whether obstacle is within width of robot + margin of error (using measuring angle, width, trig) and set the phase of the robot. New thread for obstacle following spawned when init finished (so only 1 thread spawned). 
- Main obstacle avoiding algorithm determine between calling boundary following, continuing with the previous phase, or changing to the fault phase (goal cannot be reached with current obstacle) depending on initial position when calling obstacle_avoidance and distance from robot position to goal.
- Comments provided

## Test Plan
- Tested with csv file containing sensor data and tracking result.
- Tested faulty sensor value (less than 0).
- Tested sensor value greater than range (doesn't show as obstacle). 
- Added random sensor value testing with random width.
- Tested that detection range is right with the correct width.

## Other
Math for calculating correct sensor range is done with help by Aayush and Ellie


engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+                    with open(ROOT_DIR + '/tests/functionality_tests/csv/avoid_obstacle_result.csv', 'a') as fd:
+                        fd.write("Not Avoid" + '\n')
+            if curr_ultrasonic_value < 0:
+                self.set_phase(Phase.FAULT)  # value should not go below 0; sensor is broken
+                with open(ROOT_DIR + '/tests/functionality_tests/csv/avoid_obstacle_result.csv', 'a') as fd:
+                    fd.write("Fault" + '\n')
+
+        # time.sleep(10)  # don't hog the cpu
+
+    def execute_avoid_obstacle(self, dist_to_goal, prev_phase, init_threshold, goal_threshold):
```
sazhang02, line 417: all of these parameters are initialized in the robot. Should we do self.parameter instead? esp bc self.dist_to_goal seems to be updated in track_obstacle?

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+                    goal is unreachable
+                goal_threshold (Double): Threshold from goal that will be detected as reaching goal
+            Returns:
+                prev_phase (Phase)
+        """
+        init_x = self.x_pos
+        init_y = self.y_pos
+        # threshold is arbitrary, set later. sometimes location reading will be inaccurate or not frequent enough to
+        # read that we've arrived back to the init x y pos. make it small enough that robot actually leaves threshold
+        # at some time during boundary following or add timeout to branch making gate = True
```
sazhang02, line 432: is this a todo? also which threshold, init or goal?

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+        gate = False
+        dist_to_goal_decreases = False
+        boundary_traversed = False
+        while True:
+            if not self.fault:  # fault has priority over obstacle avoidance
+                dist_from_init = math.sqrt((self.x_pos - init_x) ** 2 + (self.y_pos - init_y) ** 2)
```
sazhang02, line 438: maybe we should make a general distance helper function + also use in track_obstacle

engine/robot.py
```diff
@@ -302,26 +338,34 @@ def traverse_roomba(self, base_station_loc, time_limit, roomba_radius):
...
-            new_x = curr_x + self.move_dist * \
-                math.cos(self.state[2]) * self.time_step
-            new_y = curr_y + self.move_dist * \
-                math.sin(self.state[2]) * self.time_step
+            new_x = curr_x + self.move_dist * math.cos(self.state[2]) * self.time_step
+            new_y = curr_y + self.move_dist * math.sin(self.state[2]) * self.time_step
             next_radius = math.sqrt(
-                abs(new_x-base_station_loc[0])**2 + abs(new_y-base_station_loc[1])**2)
-            if next_radius > roomba_radius:
+                abs(new_x - base_station_loc[0]) ** 2 + abs(new_y - base_station_loc[1]) ** 2)
```
sazhang02, line 353: helper function to calculate radius?

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+        # read that we've arrived back to the init x y pos. make it small enough that robot actually leaves threshold
+        # at some time during boundary following or add timeout to branch making gate = True
+        gate = False
+        dist_to_goal_decreases = False
+        boundary_traversed = False
+        while True:
+            if not self.fault:  # fault has priority over obstacle avoidance
+                dist_from_init = math.sqrt((self.x_pos - init_x) ** 2 + (self.y_pos - init_y) ** 2)
+                if dist_from_init > init_threshold:
+                    gate = True
```
sazhang02, line 440: Maybe we can rename gate? Not sure what a good name would be here. I am also getting a lil confused about what init_threshold is again ;o;

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+            Returns:
+                prev_phase (Phase)
+        """
+        init_x = self.x_pos
+        init_y = self.y_pos
+        # threshold is arbitrary, set later. sometimes location reading will be inaccurate or not frequent enough to
+        # read that we've arrived back to the init x y pos. make it small enough that robot actually leaves threshold
+        # at some time during boundary following or add timeout to branch making gate = True
+        gate = False
+        dist_to_goal_decreases = False
```
sazhang02, line 434: has_decreased_goal_dist? is_decreasing_goal_dist? trying to make it sound like a condition

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+        boundary_traversed = False
```
sazhang02, line 435: has_traversed_boundary?

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+                    self.set_phase(Phase.FAULT)  # cannot reach goal
+                    return None
+                elif dist_to_goal_decreases:
+                    # don't include condition on whether there is a new obstacle bc it will be caught and
+                    # current algorithm will continue traversing current boundary (instead of traversing new obstacle)
+                    # not optimal because frequently will have to go back to obstacle avoidance
+                    self.set_phase(prev_phase)  # goes back to prev traversal
+                    return None
+                    # check position of goal and if obstacle in way of goal (using side sensors) then keep doing
+                    # boundary following except with new init_x, init_y, gate,
```
sazhang02, line 454: is this a todo? after a return statment

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+                if curr_dist_to_goal < goal_threshold and gate:  # exits obstacle avoidance if robot close to goal
+                    self.set_phase(prev_phase)
+                    return None
+                elif boundary_traversed and gate:
+                    self.set_phase(Phase.FAULT)  # cannot reach goal
+                    return None
+                elif dist_to_goal_decreases:
+                    # don't include condition on whether there is a new obstacle bc it will be caught and
+                    # current algorithm will continue traversing current boundary (instead of traversing new obstacle)
+                    # not optimal because frequently will have to go back to obstacle avoidance
```
sazhang02, line 450: So we want to transition back to our previous phase as long as we are closer to the goal than we were previously?

I feel like this might lead to a lot of phase transitions in and out of the obstacle avoidance phase if the obstacle is still in the detection range of the robot. This could potentially lead to concurrency issues. If my understanding is correct, we might have the following scenario?

For example, consider the case where R = robot, G = Goal, O = Obstacle, and the goal is within the detection range and each | | is a cell/node after a timestep of moving forward.
|G|
|   |
|O|
|   |
 R
If the robot proceeds forward one time step, it hasn't successfully avoided the obstacle but its closer to the goal. So, we would switch to the previous phase (lets say traverse for example).
|G|
|   |
|O|
|R|
There could be a case where move_to_target_node begins executes before track_obstacle "interrupts" the traversal phase. Then collision would happen?
Or do we assume `dist_to_goal_decreases` being true means boundary following takes the robot out of the way from the obstacle first?

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
...
+                    self.execute_boundary_following(0)  # add code directly here
+                    # update conditions
+                    curr_dist_to_goal = math.sqrt(
+                        (self.x_pos - self.goal_location[0]) ** 2 + (self.y_pos - self.goal_location[1]) ** 2)
+                    dist_to_goal_decreases = (curr_dist_to_goal < dist_to_goal)
+                    boundary_traversed = dist_from_init < init_threshold
+            else:
+                self.set_phase(Phase.FAULT)
+                return None
+            time.sleep(10)  # don't hog the cpu
```
sazhang02, line 465: This makes sense in track_obstacle but why time sleep here

tests/functionality_tests/obstacle_tracking_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+class TestObstacleTracking(unittest.TestCase):
+
+    def populate_standard_obstacle_data(self):
+        mock_data_path = ROOT_DIR + '/tests/functionality_tests/csv/ultrasonic_values.csv'
+        mock_data_file = open(mock_data_path, "w")
+        mock_data_file.truncate()
+        mock_data_file.write(
+            "(600, 600, 600, 600, 600, 'assuming a straight line between robot and goal with rectangular box centered in the middle')\n")
```
sazhang02, line 18: maybe add a comment above w/ what each csv parameter represents for readability? like (front_value,.....)

tests/functionality_tests/obstacle_tracking_test.py
```diff
@@ -0,0 +1,195 @@
...
+            lf_value = random.random() * 600
+            lb_value = random.random() * 600
+            rf_value = random.random() * 600
+            rb_value = random.random() * 600
+            data = "(" + str(front_sensor_value) + ", " + str(lf_value) + ", " + str(lb_value) + ", " + str(
+                rf_value) + ", " + str(rb_value) + ")"
+            mock_data_file.write(data + "\n")
+        mock_data_file.close()
+
+    def approximate(self, threshold, val1, val2):
```
sazhang02, line 100: maybe rename to be more specific about what we are approximating?

tests/functionality_tests/obstacle_tracking_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def test_detection_range(self):
+        wide_robot = Robot(0, 0, math.pi / 2, epsilon=0.2, max_v=0.5, radius=0.2, init_phase=2, width=800,
+                           front_ultrasonic=None, lb_ultrasonic=None, lf_ultrasonic=None, rb_ultrasonic=None,
+                           rf_ultrasonic=None)
+        skinny_robot = Robot(0, 0, math.pi / 2, epsilon=0.2, max_v=0.5, radius=0.2, init_phase=2, width=200,
+                             front_ultrasonic=None, lb_ultrasonic=None, lf_ultrasonic=None, rb_ultrasonic=None,
+                             rf_ultrasonic=None)
+        self.assertEqual(self.approximate(1, wide_robot.max_sensor_range, wide_robot.detect_obstacle_range), True)
+        self.assertEqual(self.approximate(1, 165.1, skinny_robot.detect_obstacle_range), True)
```
sazhang02, line 191: Do any false cases exist

sazhang02 (commented):
YAYYY obstacle avoidance algorithm!! \o////// It is looking very good. Just some minor things and a few questions I have.

sazhang02 (changes requested):
oops i forgot to hit request changes

engine/robot.py
```diff
@@ -94,14 +103,35 @@ def __init__(self, x_pos, y_pos, heading, epsilon, max_v, radius, is_sim=True, p
...
         self.plastic_weight = plastic_weight
         self.battery = 100  # TEMPORARY
         self.acceleration = [0, 0, 0]  # TEMPORARY
         self.magnetic_field = [0, 0, 0]  # TEMPORARY
         self.gyro_rotation = [0, 0, 0]  # TEMPORARY
         self.linear_v = 0
         self.angular_v = 0
+        self.width = width
+        self.avoid_obstacle = False
+        self.fault = False
```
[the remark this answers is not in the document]
AC408: width should be in documentation (of params) alr

engine/robot.py
AC408: we don't assume that dist_to_goal_decreases being true means boundary following takes the robot out of the way of the obstacle
To fix the concurrency issue, we can have another function where the robot turn towards the target and then check and then potentially obstacle avoiding again?

engine/robot.py
AC408: so it allows the other (track_obstacle) thread to run. Otherwise, the cpu will never be allocated to the other thread... i think?

sazhang02: ok ye makes sense! we can test out the sleep time later? i feel like 10 might be a bit long

engine/robot.py
sazhang02: Where would this check be?

sazhang02:
psst also the hidden comments

engine/robot.py
AC408: We can put this in a separate function that will be called (within the avoid obstacle algorithm) before returning to the traversal phase

engine/robot.py
AC408: yeah, we can test it later

engine/robot.py
```diff
@@ -302,26 +338,34 @@ def traverse_roomba(self, base_station_loc, time_limit, roomba_radius):
             Returns:
                 None
         """
+        # to add obstacle avoidance, can either change roomba to go to point on the circle in the direction of the
+        # robot direction or use raw sensor value
+        # can we guarantee that when giving a velocity that the robot will move at that velocity?
```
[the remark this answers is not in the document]
AC408: comments removed.

engine/robot.py
AC408: we need dist_to_goal to stay static, so we'll be taking it from the parameter. The other params are removed

engine/robot.py
AC408: It's just a note/reasoning for my implementation. Threshold is for init_thresholds; updated.

engine/robot.py
sazhang02: YUS

engine/robot.py
AC408: i just used generic calculate_dist function

engine/robot.py
AC408: did_dist_to_goal_decreased

engine/robot.py
AC408: note/reasoning about how algorithm would rework. Will be redone with the below idea about having another function check before going to prev phase/continuing to do obst avoid

engine/robot.py
```diff
@@ -330,8 +374,97 @@ def set_phase(self, new_phase):
         with open(CSV_PATH + '/phases.csv', 'a') as fd:
             fd.write(str(self.phase) + '\n')
 
-    def execute_avoid_obstacle(self):
-        # TODO: SET BACK TO ORIGINAL MISSION (TRAVERSE OR RETURN)
+    def track_obstacle(self):
+        # make this async so main algorithm keeps running
+        # assuming front sensor is mounted in the center x position (width/2)
+        condition = True
```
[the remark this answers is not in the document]
AC408: This condition was added because of testing. If its the real robot, we want to continuously check for obstacles. If its a simulation, we dont want a while loop because there's a finite mock sensor data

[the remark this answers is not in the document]
AC408: not sure how i would rename it beside smthg like "want_to_get_sensor_data" or smthg

engine/robot.py
AC408: for gate, what about "was_out_of_init_threshold"? I dont know how to rename init_threshold while keeping it short.

engine/robot.py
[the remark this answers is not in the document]
sazhang02: so if its on our real robot it should be True and otherwise False?

Maybe we can use robot.is_sim?

sazhang02:
answered comment

engine/robot.py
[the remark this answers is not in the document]
AC408: making try except and reducing condition to true

engine/robot.py
AC408: has_left_init_thresh

sazhang02 (changes requested):

tests/functionality_tests/csv/avoid_obstacle_result.csv
```diff
@@ -0,0 +1,10 @@
+Not Avoid
+Not Avoid
+Avoid
+Not Avoid
+Avoid
+Not Avoid
+Not Avoid
+Not Avoid
+Not Avoid
+Not Avoid
```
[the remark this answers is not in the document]
AC408: we dont need to keep the data. It's written when the test was ran.

tests/functionality_tests/csv/ultrasonic_values.csv
```diff
@@ -0,0 +1,10 @@
+(259.9502880289185, 94.54772172034791, 365.8677760854979, 175.7180851320918, 4.552129332283328)
+(126.35092952516614, 72.77332120196715, 45.862415872772644, 393.8322736277804, 440.56132938911344)
+(79.14528372742183, 27.672808583937236, 182.78482577488302, 338.78106568582024, 520.9234641692442)
+(621.3997581786865, 14.594753751851819, 144.3092004204622, 186.06506306910293, 167.3661657042625)
+(79.71341785769256, 427.56864012654455, 411.4795856221073, 373.52226959291585, 478.5436795696814)
+(474.5346026848112, 122.92711853689806, 335.6464119178094, 103.9696823119313, 365.4474955457613)
+(467.4669394024443, 230.7166132860191, 425.02668308595423, 82.41854180799544, 81.21606703700897)
+(230.61915427522004, 307.11538384054, 25.52458815790386, 166.96218628662746, 41.35512568290209)
+(474.9126347640701, 198.36533421988668, 179.97414508193066, 271.75160636566375, 292.75778846480813)
+(119.79293820312863, 513.8906463031104, 63.43467720764853, 195.66474767234942, 376.62631080660884)
```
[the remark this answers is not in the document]
AC408: we dont need to keep the data. It's written when the test was ran.

tests/functionality_tests/csv/avoid_obstacle_result.csv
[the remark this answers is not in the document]
sazhang02: delete?

tests/functionality_tests/csv/ultrasonic_values.csv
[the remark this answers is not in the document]
sazhang02: also delete?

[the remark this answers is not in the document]
sazhang02: + we can gate the csv write once the other pr goes through 🤔 doesn't have to be now tho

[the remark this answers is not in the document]
AC408: yee

sazhang02 (approved):
SHOULD BE GOOD \O/

Aayush-Agnihotri (approved):
Looks good!

Ellied5 (approved):

PR #97, opened 2022-10-29
Is raspberrypi

jinlin270:
## References
[stack exchange (where the code came from)](https://raspberrypi.stackexchange.com/questions/5100/detect-that-a-python-program-is-running-on-the-pi/74541#74541)

### Description

Copy the is_raspberrypi function from StackExchange and confirmed it works on MacOS and raspberrypi. 

## Test Plan

Tested on  MacOS and raspberrypi. May test on a Window and Linux afterwards.
## Other

The next step will be integrating the function to automate is_sim flag.


engine/grid.py
```diff
@@ -206,14 +206,14 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
+    def activate_rectangle(self, row, col, row_limit, col_limit):
         """
         Activates all the nodes in a rectangle.
         """
         for x in range(row, row_limit):
             for y in range(col, col_limit):
-                nodes[x,y].is_active = True
-        return nodes
+                self.nodes[x,y].is_active = True
+        return self.nodes
```
sazhang02, line 217: hmm can we take these changes out of this pr? i think its from the grid refactoring branch?

engine/is_raspberrypi.py
```diff
@@ -0,0 +1,13 @@
+import io
+import os
+"""a function from : https://raspberrypi.stackexchange.com/questions/5100/detect-that-a-python-program-is-running-on-the-pi/74541#74541
+return True if the current device is a raspberry_pi, False otherwise"""
+def is_raspberrypi():
+    try:
```
sazhang02, line 6:
```suggestion
def is_raspberrypi():
    """
    Return True if the current device is a raspberry_pi, False otherwise.
    Reference: https://raspberrypi.stackexchange.com/questions/5100/detect-that-a-python-program-is-running-on-the-pi/74541#74541
    """
    try:
```

sazhang02: oops i dont think this suggestion will work bc its not recognizing the multiline comment, but think we can move the spec under the function header + I changed a function from to Reference

sazhang02 (changes requested):
Nice! Suggested some small changes

sazhang02 (approved):
YAY

PR #98, opened 2022-11-05
Fix Lawnmower to not zigzag at sharp angle

jinlin270:
## Summary

This PR updates the lawnmower algorithm to go directly up and traverse backward at the end of each row instead of going to the start of the row and traversing right each time. I also updated the graphs in src/tests/functionality_tests/activate_shape_graphs to reflect this new traversal. 

### Description

The new traversal will reduce the amount of distance the robot traverses, and make all the turning angles 90° so the robot can turn more easily.

This is what the lawnmower looks like now:
Screen Shot 2022-11-04 at 8 18 55 PM

this is what the lawnmower traversal looks like before:
Screen Shot 2022-11-04 at 8 17 37 PM

## Test Plan

The graphs in test_activation_shapes :)


engine/grid.py
```diff
@@ -303,70 +304,106 @@ def find_border_nodes(self):
...
         leftmost_node = None
+        rightmost_node = None
         leftmost_node_pos = None
+        rightmost_node_pos = None
+        
         for row in range(rows):
             for col in range(cols):
                 node = self.nodes[row][col]
+                #border node 1 left active node 
                 if node.is_active and self.is_on_border(row, col, rows-1, cols-1):
```
juliehelenek, line 315: let's add documentation / a comment explaining how this if statement is different from the node below it, i.e "Checking if current node is active and is on the left side of the border" or something

engine/grid.py
```diff
@@ -303,70 +304,106 @@ def find_border_nodes(self):
...
                     # check if this is an active node and on the border
                     self.nodes[row][col].is_border = True
                     border_list.append((node, row, col))
                     if leftmost_node_pos is None or col < leftmost_node_pos[1]:
                         leftmost_node = node
                         leftmost_node_pos = (row, col)
+                if node.is_active and self.is_on_border(row, col, rows , cols):
+                    # check if this is an active node and on the border
+                    self.nodes[row][col].is_border = True
+                    if rightmost_node_pos is None or row > rightmost_node_pos[0]:
```
juliehelenek, line 325: if the current node has a row attribute i think changing this to 
if rightmost_node_pos is None or node.row > rightmost_node_pos[0] would be a little clearer, but not the biggest deal

engine/grid.py
```diff
@@ -303,70 +304,106 @@ def find_border_nodes(self):
...
-
         # node_info: (node, row, col)
         candidate_nodes = [node_info for node_info in self.border_nodes if node_info[2] == pos[1]+1]
         if (candidate_nodes == []):
             return None
         else:
             node_info = min(candidate_nodes,key=lambda node_info: node_info[1])
             return (node_info[1],node_info[2])
+    
+    def bottom_leftmost_node(self, pos):
```
juliehelenek, line 348: Let's make sure these have proper documentation; 
def function_name(self, parameter1, parameter2):
   """
   Short description of function
   
   Parameters: 
   parameter1: short description 
   precondition: does parameter 1 have to be a certain value?
   
   parameter2: short description
   precondition: does parameter 2 have to be a certain value?
   
   returns: (or describe procedure if it doesnt return anything)

   """

engine/grid.py
```diff
@@ -303,70 +304,106 @@ def find_border_nodes(self):
...
         waypoints.append(curr_pos)
         while (phase != WaypointPhase.TERMINATE):
-            if (phase == WaypointPhase.DOWN):
-                new_pos = (curr_pos[0]+1, curr_pos[1])
-                if curr_pos[0]+1 == rows and self.nodes[new_pos].is_active:
+            if direction == Direction.LEFT:
+                new_pos = (curr_pos[0]-1,curr_pos[1])
+                # next_next_pos = (curr_pos[0]-2,curr_pos[1])
+                if self.nodes[new_pos].is_active:
+                    print("branch 2 ",new_pos)
```
juliehelenek, line 377: lets remove this print!

engine/grid.py
```diff
@@ -303,70 +304,106 @@ def find_border_nodes(self):
...
                     waypoints.append(new_pos)
-                    right_pos = self.bottom_rightmost_node(new_pos)
-                    if right_pos is not None:
-                        waypoints.append(right_pos)
-                        curr_pos = right_pos
-                        phase = WaypointPhase.DOWN
+                    curr_pos = new_pos
+                else:
+                    print("branch 3",new_pos)
```
juliehelenek, line 381: remove print!

engine/grid.py
```diff
@@ -303,70 +304,106 @@ def find_border_nodes(self):
...
+                        waypoints.append(left_pos)
+                        curr_pos = left_pos
+                        direction = Direction.RIGHT
                     else:
                         phase = WaypointPhase.TERMINATE
-                elif self.nodes[new_pos].is_active:
+            if direction == Direction.RIGHT:
+                new_pos = (curr_pos[0]+1,curr_pos[1])
+                if self.nodes[new_pos].is_active:
+                    print("branch 2", new_pos)
```
juliehelenek, line 392: remove!

engine/grid.py
```diff
@@ -303,70 +304,106 @@ def find_border_nodes(self):
...
                     waypoints.append(new_pos)
                     curr_pos = new_pos
                 else:
-                    right_pos = self.bottom_rightmost_node(new_pos)
+                    print("branch 3", new_pos)
```
juliehelenek, line 396: remove!

engine/grid.py
```diff
@@ -515,9 +552,10 @@ def get_waypoints(self, mode):
...
             waypoints = self.get_spiral_waypoints()
         elif mode == ControlMode.STRAIGHT:
-            waypoints = self.get_straight_line_waypoints(y_start_pct=0.5)
+            waypoints = self.get_straight_line_waypoints(y_start_pct =0.5)
         elif mode == ControlMode.LAWNMOWER_A:
             waypoints = self.get_all_lawnmower_waypoints_adjustable()
         else:
             return []
         return waypoints
+
```
juliehelenek, line 561: im putting this comment here since i cant add it to circle.png: there seem to be some sharp turns still - im assuming that the sock traversal isnt part of this PR? if it is not part of the PR, then nvm, ignore this comment!!

juliehelenek (changes requested):

engine/grid.py
jinlin270: I forgot what this comment is, but I think the if statement is explained in the comment below? I can delete this comment since it is unused

engine/grid.py
jinlin270: Actually the row is not an attribute, we are just iterating from 0 to the total number of rows. If this is confusing we can name it something else?

engine/grid.py
jinlin270: does this look right?
"""
        A helper function for get_all_lawnmower_waypoints_adjustable, finds the leftmost
        node of the row above current pos.

        Parameters:
        parameter1: pos
        precondition: None

        returns:  Returns the furthest left node on the row above the current pos, 
        or return None if there is no nodes in the row above.
        """

engine/grid.py
jinlin270: Ooops, all prints removed now

engine/grid.py
jinlin270: It is not a part of this PR

engine/grid.py
jinlin270: Aye Aye

engine/grid.py
jinlin270: Aye Aye

engine/grid.py
jinlin270: Aye Aye

engine/grid.py
jinlin270: Yes

engine/grid.py
jinlin270: It is not a part of the PR

engine/grid.py
jinlin270: Actually row is not an attribute, we are iterating from 0 to rows, and row is the counter :D

engine/grid.py
jinlin270: The if statement is documented in a comment below, I have no idea what this comment is about. I will go delete it

engine/grid.py
jinlin270: does this look right?

        """
        A helper function for get_all_lawnmower_waypoints_adjustable, finds the leftmost
        node of the row above current pos.

        Parameters:
        parameter1: pos
        precondition: None

        returns:  Returns the furthest left node on the row above the current pos, 
        or return None if there is no nodes in the row above.
        """

engine/grid.py
```diff
@@ -314,59 +318,109 @@ def find_border_nodes(self):
                     if leftmost_node_pos is None or col < leftmost_node_pos[1]:
                         leftmost_node = node
                         leftmost_node_pos = (row, col)
+                if node.is_active and self.is_on_border(row, col, rows , cols):
+                    # check if this is an active node and on the border
+                    self.nodes[row][col].is_border = True
+                    if rightmost_node_pos is None or row > rightmost_node_pos[0]:
+                        rightmost_node = node
+                        rightmost_node_pos = (row, col)
```
sazhang02, line 326: why is this change necessary? when i comment it out it runs alright

engine/grid.py
```diff
@@ -314,59 +318,109 @@ def find_border_nodes(self):
...
+
+        returns:  Returns the furthest left node on the row above the current pos, 
+        or return None if there is no nodes in the row above.
+        """
+        # node_info: (node, row, col)
+        candidate_nodes = [node_info for node_info in self.border_nodes if node_info[2] == pos[1]+1]
+        if (candidate_nodes == []):
+            return None
+        else:
+            node_info = max(candidate_nodes,key=lambda node_info: node_info[1])
```
sazhang02, line 375: It seems like this is the only change between bottommost left and right. What do you think about combining those two functions into one general function such as
`next_row_starting_node(self, pos, current_direction)`

Rather than bottommost left/right looking at ur spec and the code again, the code gives you the next node in the row above you want to traverse to once you reach an edge. We could take Direction(Enum) outside of get all lawnmower and send it in as a parameter to determine whether you should min/max

sazhang02: This also made me realize that we should eventually expand between rowbase traversal and colummbase traversal since rn our current lawnmower impl used in sim_trajectory is column based

If we wanna name the above, we could likewise call the column-base equivalent next_column_starting_node...

engine/grid.py
```diff
@@ -226,70 +228,69 @@ def activate_circle(self, circle_center_row, circle_center_col, circle_radius):
...
+    ##Checks if node is on border of activated nodes
+    def is_on_border(self,row, col, row_limit, col_limit):
+            min_col = max(0, col-1)
+            min_row = max(0, row-1)
+            max_col = min(col_limit, col+1)
+            max_row = min(row_limit, row+1)
+
+            # If this node is on the very edge of the grid, it is automatically a border node
+            if min_col == 0 or min_row == 0 or max_col == col_limit or max_row == row_limit:
+                return True
```
sazhang02, line 284: Thinking about this bc we have double borders if we try to use almost all the grid points (start row/col = 0, end row/col = rows/cols -1)
I tried
```
        if row == 0 or col == 0 or col == self.get_num_cols or row == self.get_num_rows:
            return True
```
and it fixed the double borders but there is a problem w the circle border (see pics below)

I think work on this can be in a separate PR, unless you think this could affect the current impl for lawnmower!

engine/grid.py
```diff
@@ -314,59 +318,109 @@ def find_border_nodes(self):
...
+        direction = Direction.RIGHT
         waypoints = []
         waypoints.append(curr_pos)
         while (phase != WaypointPhase.TERMINATE):
-            if (phase == WaypointPhase.DOWN):
-                new_pos = (curr_pos[0]+1, curr_pos[1])
-                if curr_pos[0]+1 == rows and self.nodes[new_pos].is_active:
+            if direction == Direction.LEFT:
+                new_pos = (curr_pos[0]-1,curr_pos[1])
+                # next_next_pos = (curr_pos[0]-2,curr_pos[1])
```
sazhang02, line 396: delete

engine/grid.py
sazhang02: Before
Screen Shot 2022-11-14 at 2 53 14 AM
Screen Shot 2022-11-14 at 2 53 30 AM

After
Screen Shot 2022-11-14 at 2 52 25 AM
Screen Shot 2022-11-14 at 2 53 43 AM

sazhang02 (changes requested):
Nice! Some comments + question about find border nodes

AdityaKompella6 (approved):
This one is all good. No zig zag = good.

juliehelenek (approved):
lin addressed all my comments! sammie had a few that still could be done ie refactoring code!

PR #103, opened 2022-11-12
automate is_sim flag

jinlin270:
## Summary
Automate is_sim flag in robot.py, sim_trajectory.py, and gui.py. 

### Description

is_sim flag is True when the code is running on a simulation and false otherwise. Before this flag is manually set, now this PR automates the flag using is_raspberrypi.py which automatically check is the code is running on a raspberry pi. If the code is running on a raspberry pi, the is_sim flag is False. 

## Test Plan

check these simulation still works: 
python -m gui.gui 
python -m engine.sim_trajectory


sazhang02 (approved):
Great!

PR #104, opened 2022-11-12
Sock traversal

jinlin270:
## Summary
This PR updates the traversal algorithm to turn in a smoother, circular motion. The graphs in src/tests/functionality_tests/activate_shape_graphs are also updated to reflect this new traversal. A traversal test for a vertical line is added to activate_shape_test to cover this edge case.

This is what the lawnmower looks like now:
Screen Shot 2022-11-12 at 2 39 42 PM
Screen Shot 2022-11-12 at 2 40 08 PM
Screen Shot 2022-11-12 at 2 40 30 PM
Screen Shot 2022-11-12 at 2 40 41 PM

### Description

The new traversal will avoid turning over 90 degrees.

## Test Plan

run python3 -m tests.functionality_tests.test_activation_shapes
and input one of : circle ; rectangle ; triangle ; line

## Other

Allen did the cool circle traversal.


engine/grid.py
```diff
@@ -115,14 +121,16 @@ def generate_nodes(start_lat, start_long, rows, cols, step_size_m):
...
         self.active_waypoints_list = []
         self.inactive_waypoints_list = []
 
     def get_active_waypoints_list(self):
         return self.active_waypoints_list
-
+    
     def get_inactive_waypoints_list(self):
-        return (self.inactive_waypoints_list)
+        return(self.inactive_waypoints_list)
```
AC408, line 133:
```suggestion
        return self.inactive_waypoints_list
```

engine/grid.py
```diff
@@ -183,6 +192,7 @@ def is_on_border(self, row, col, row_limit, col_limit):
                     return True
         return False
 
+
     # --------------------- ADJUSTABLE TRAVERSAL ALGORITHMS -------------- #
 
     def get_neighbor_node(self, row, col, row_max, col_max):
```
AC408, line 198:
```suggestion
    def get_active_neighbor_node(self, row, col, row_max, col_max):
```

engine/grid.py
```diff
@@ -200,96 +210,93 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
+            # If this node is on the very edge of the grid, it is automatically a border node
+            if min_col == 0 or min_row == 0 or max_col == col_limit or max_row == row_limit:
+                return True
+
+            # If the node has a neighboring node that is inactive, it is a border node
+            for col in range(min_col, max_col+1):
+                for row in range(min_row, max_row+1):
+                    if not self.nodes[row][col].is_active:
+                        return True
+            return False
```
AC408, line 298: isn't this only checking if one node is on the border of activated nodes? why is there a loop?

engine/grid.py
```diff
@@ -200,96 +210,93 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
         A3 = area(x1, y1, x2, y2, x, y)
-
+        
         # Check if sum of A1, A2 and A3
         # is same as A
-        if (A == A1 + A2 + A3):
+        if(A == A1 + A2 + A3):
             return True
         else:
             return False
```
AC408, line 264:
```suggestion
        return A == (A1 + A2 + A2)
```

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
         leftmost_node_pos = None
-        for row in range(rows):
-            for col in range(cols):
+        rightmost_node_pos = None
+        
+        for row in range(self.num_rows):
+            for col in range(self.num_cols):
                 node = self.nodes[row][col]
-                if node.is_active and self.is_on_border(row, col, rows-1, cols-1):
+                #border node 1 left active node
```
AC408, line 318: what does this comment mean?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
                     border_list.append((node, row, col))
                     if leftmost_node_pos is None or col < leftmost_node_pos[1]:
                         leftmost_node = node
                         leftmost_node_pos = (row, col)
+                if node.is_active and self.is_on_border(row, col, self.num_rows , self.num_cols):
+                    # check if this is an active node and on the border
+                    self.nodes[row][col].is_border = True
+                    if rightmost_node_pos is None or row > rightmost_node_pos[0]:
+                        rightmost_node = node
+                        rightmost_node_pos = (row, col)
```
AC408, line 331: why is one of them passing self.num_rows and the other one self.num_rows-1? Also, it looks like both if expressions can be combined

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
+        for row in range(self.num_rows):
+            for col in range(self.num_cols):
                 node = self.nodes[row][col]
-                if node.is_active and self.is_on_border(row, col, rows-1, cols-1):
+                #border node 1 left active node 
+                if node.is_active and self.is_on_border(row, col, self.num_rows-1, self.num_cols-1):
                     # check if this is an active node and on the border
                     self.nodes[row][col].is_border = True
                     border_list.append((node, row, col))
                     if leftmost_node_pos is None or col < leftmost_node_pos[1]:
```
AC408, line 323: when updating leftmost_node_pos and leftmost_node, why are we concerned about the col and not the row?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
             return (node_info[1],node_info[2])
+    
+    def bottom_leftmost_node(self, pos):
+        # node_info: (node, row, col)
+        candidate_nodes = [node_info for node_info in self.border_nodes if node_info[2] == pos[1]+1]
+        if (candidate_nodes == []):
+            return None
+        else:
+            node_info = max(candidate_nodes,key=lambda node_info: node_info[1])
+            return (node_info[1],node_info[2])
```
AC408, line 359: can be refactored into one function and an additional param. Also, by bottom most, is this referring to the bottom most of the entire grid/shape or the current col?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
+            theta_init = math.atan2(start_pos[1]-center[1], start_pos[0]-center[0])
+            theta_end = math.atan2(end_pos[1]-center[1], end_pos[0]-center[0])
+            circle_plt = []
+            if theta_end < theta_init:
+                theta_end = theta_end + 2*math.pi
+            if orientation == Orientation.CCW:
+                theta = theta_init
+                while theta < theta_end:
+                    circle_plt.append((r*math.cos(theta)+center[0], r*math.sin(theta)+center[1]))
+                    theta = theta + math.pi/12
```
AC408, line 399: can be made into param with default value math.pi/12 to give user customizability to how smooth the curve is

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
+                if self.nodes[new_pos].is_active:
                     waypoints.append(new_pos)
-                    right_pos = self.bottom_rightmost_node(new_pos)
-                    if right_pos is not None:
-                        waypoints.append(right_pos)
-                        curr_pos = right_pos
-                        phase = WaypointPhase.DOWN
+                    curr_pos = new_pos
+                else:
+                    left_pos = self.bottom_rightmost_node(new_pos)
```
AC408, line 421: does bottom_rightmost_node of the new_position return the left_pos? I'm confused by the naming and what bottom_rightmost_node does

AC408: also, it looks like bottom_rightmost_node return the right-most activated node of the row directly above the current row, is that right?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
                     # check if this is an active node and on the border
                     self.nodes[row][col].is_border = True
                     border_list.append((node, row, col))
                     if leftmost_node_pos is None or col < leftmost_node_pos[1]:
                         leftmost_node = node
                         leftmost_node_pos = (row, col)
+                if node.is_active and self.is_on_border(row, col, self.num_rows , self.num_cols):
+                    # check if this is an active node and on the border
+                    self.nodes[row][col].is_border = True
+                    if rightmost_node_pos is None or row > rightmost_node_pos[0]:
```
AC408, line 329: when updating rightmost_node_pos and rightmost_node, why are we concerned about the row and not the col?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
-                        curr_pos = right_pos
-                        phase = WaypointPhase.DOWN
+                    curr_pos = new_pos
+                else:
+                    left_pos = self.bottom_rightmost_node(new_pos)
+                    if left_pos is not None:
+                        new_pos = (new_pos[0]+1, new_pos[1])
+                        while new_pos[0] < left_pos[0]:
+                            while new_pos in waypoints:
+                                waypoints.remove(new_pos)
```
AC408, line 426: isn't this removing your own progress? On line 423, isn't new_pos curr_pos (since we have new_pos = (curr_pos[0]-1,curr_pos[1]) on line 416). In that case, if you're going back to the right, aren't you removing the waypoints you just added?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
+                            new_pos = (new_pos[0]+1, new_pos[1])
+                        circle_plt = plot_circle((left_pos[0], left_pos[1]), (left_pos[0], left_pos[1]-1),
+                                                 (left_pos[0], left_pos[1]-.5), Orientation.CW)
+                        waypoints += circle_plt
+                        direction = Direction.RIGHT
+                        if self.nodes[(left_pos[0]+1,left_pos[1])].is_active:
+                            curr_pos = left_pos
```
AC408, line 433: isn't this guaranteed to be false since left_pos is the rightmost activated node, so anything to its right will never be active?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
+                        elif self.bottom_rightmost_node((new_pos[0],new_pos[1]+1)) is not None :
+                            curr_pos = self.bottom_rightmost_node((new_pos[0],new_pos[1]))
```
AC408, line 435: what does this do?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
+            node_info = max(candidate_nodes,key=lambda node_info: node_info[1])
+            return (node_info[1],node_info[2])
 
+    
 
-    # Given border and active nodes, compute lawnmower traversal
-
+        
+    ##Given border and active nodes, compute lawnmower traversal
     def get_all_lawnmower_waypoints_adjustable(self):
```
AC408, line 365: documentation on how its being traversed? (left to right then top to bottom, for ex, (7, 0) -> (7, 9) -> (0,0)

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
-        rows = self.nodes.shape[0]
+        class Direction(Enum):
+            RIGHT = 1
+            LEFT = 2
+
+        class Orientation(Enum):
+            CW = 1
+            CCW = 2
+
+        def plot_circle(start_pos, end_pos, center, orientation):
```
AC408, line 377: has this been tested?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
+                    curr_pos = new_pos
+                else:
+                    left_pos = self.bottom_rightmost_node(new_pos)
+                    if left_pos is not None:
+                        new_pos = (new_pos[0]+1, new_pos[1])
+                        while new_pos[0] < left_pos[0]:
+                            while new_pos in waypoints:
+                                waypoints.remove(new_pos)
+                            new_pos = (new_pos[0]+1, new_pos[1])
+                        circle_plt = plot_circle((left_pos[0], left_pos[1]), (left_pos[0], left_pos[1]-1),
```
AC408, line 428: based on the documentation, this doesn't seem like it'll work. assuming the traversal is top to bottom, this would (based on the documentation), make an inverted curve?

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
                     if right_pos is not None:
-                        waypoints.append(right_pos)
-                        curr_pos = right_pos
-                        phase = WaypointPhase.DOWN
+                        while new_pos[0] > right_pos[0]:
+                            while new_pos in waypoints:
+                                waypoints.remove(new_pos)
+                            new_pos = (new_pos[0]-1, new_pos[1])
+                        circle_plt = plot_circle((right_pos[0], right_pos[1]-1), (right_pos[0],right_pos[1]),
+                                                 (right_pos[0], right_pos[1]-.5), Orientation.CCW)
```
AC408, line 454: if the traversal is top to bottom, the start and end pos seemed flipped.

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
+                        waypoints += circle_plt
+                        direction = Direction.LEFT
+                        if self.nodes[(right_pos[0]-1,right_pos[1])].is_active:
+                            curr_pos = right_pos
+                        elif self.bottom_leftmost_node((new_pos[0],new_pos[1]+1)) is not None:
+                            curr_pos = self.bottom_leftmost_node((new_pos[0],new_pos[1]))
```
AC408, line 460: same comments as for left_pos

engine/grid.py
```diff
@@ -300,73 +307,164 @@ def find_border_nodes(self):
...
-
         # node_info: (node, row, col)
         candidate_nodes = [node_info for node_info in self.border_nodes if node_info[2] == pos[1]+1]
         if (candidate_nodes == []):
             return None
         else:
             node_info = min(candidate_nodes,key=lambda node_info: node_info[1])
             return (node_info[1],node_info[2])
+    
+    def bottom_leftmost_node(self, pos):
```
AC408, line 352: is there/do we need tests?

AC408 (changes requested):
looks good so far!

engine/grid.py
jinlin270: ooops I thought I removed it, I think it is no longer necessary. I forgot what it means

engine/grid.py
jinlin270: You are right, I removed the second if branch

engine/grid.py
jinlin270: I think this function is not referenced anywhere. Allen, can you help double-check this? If so we can safely change the name or remove the function. Not sure what it can be used for.

engine/grid.py
jinlin270: I made it into a param, but I am not sure if the naming and description are accurate. does `smoothness: int representing how subdivided the angle will be when plotting the turning arch. The 
                greater the value, the smoother the curve will be. Default value = 12.` sound right?

engine/grid.py
jinlin270: Yes, that is right. I have combined bottom_rightmost_node and bottom_rightmost_node and renamed the new function nextrow_sidemost_node

engine/grid.py
jinlin270: It seems like my code is different, I have this
`  def find_border_nodes(self):
        """
            Find all activated border nodes on the grid.

            This function loops through all the nodes, checks if a particular node has
            been activated, and if so checks to see if that node is a border node. At the
            end of the function call, fields 'leftmost_node', 'leftmost_node_pos', and
            'border_nodes' will be initialized.
        """
        border_list = []
        leftmost_node = None
        rightmost_node = None
        leftmost_node_pos = None
        rightmost_node_pos = None
        
        for row in range(self.num_rows):
            for col in range(self.num_cols):
                node = self.nodes[row][col]
                if node.is_active and self.is_on_border(row, col, self.num_rows-1, self.num_cols-1):
                    # check if this is an active node and on the border
                    self.nodes[row][col].is_border = True
                    border_list.append((node, row, col))
                    if leftmost_node_pos is None or col < leftmost_node_pos[1]:
                        leftmost_node = node
                        leftmost_node_pos = (row, col)
                    
                    
                        
        self.border_nodes = border_list
        self.leftmost_node = leftmost_node
        self.leftmost_node_pos = leftmost_node_pos
        self.rightmost_node = rightmost_node
        self.rightmost_node_pos = rightmost_node_pos`

jinlin270: However, the rightmost_node and rightmost_node_pos are never redefined and never used later. I think I added it during the zigzag traversal, but since we always traverse starting from the left side now, I can remove those attributes from the class

engine/grid.py
jinlin270: Yes, but this is necessary to remove any over-traversal when moving up to a row with fewer nodes. If we do not remove those nodes, the graph will stick out like this
Screen Shot 2022-11-13 at 7 50 49 PM

engine/grid.py
jinlin270: It is not true for the line case. 
Screen Shot 2022-11-13 at 7 56 27 PM

engine/grid.py
jinlin270: This is specifically added for the line case, so we do not terminate if we still have more rows left to traverse.

engine/grid.py
jinlin270: I don't believe the plot_circle has been unit_tested?

engine/grid.py
jinlin270: The traversal is from bottom to top. I added the (correct?) spec to get_all_lawnmower_waypoints_adjustable

engine/grid.py
jinlin270: I don't think we have unit test for anything. The only check is the tests.functionality_tests.test_activation_shapes  we have been doing which draws the graphs

jinlin270: It will be a great future step to do more tests because there are probably more bugs

engine/grid.py
jinlin270: I refactored it. It refers to the leftmost node of the row above. Really bad naming.

engine/grid.py
jinlin270: Just did. Can you check my spec to make sure it looks alright? Thank you!

engine/grid.py
AC408: well, the second if branch is to check the rightmost_node_pos. i feel like it might be necessary if the first branch is necessary

engine/grid.py
AC408: i think it'll be better to use something like theta_step and have the user pass in pi/12 instead of 12

engine/grid.py
AC408: it looks good to me

engine/grid.py
AC408: i dont think the code i wrote is right. i'll take a look sometimes later. Current expected functionality is that it starts on the left_pos and draw a circular shape cw down towards the node directly under left_pos, but it should be starting at the bottom node and going to the top

engine/grid.py
AC408: got it

engine/grid.py
AC408: same as above, ill take a look at it

engine/grid.py
```diff
@@ -412,16 +503,14 @@ def get_all_lawnmower_waypoints(self):
...
+        for i in range(self.num_cols):
+            for j in range(self.num_rows):
                 if i % 2 == 0:
                     node = node_list[j, i]
                     self.determine_active_waypoints(node)
                     waypoints.append(node)
                 elif i % 2 == 1:
-                    row_index = rows - (j + 1)
+                    row_index = self.num_rows - (j + 1)
                     node = node_list[row_index, i]
```
AC408, line 514: comment saying this is to add the nodes in reverse order (to not zigzag). I keep forgetting why this is what it is

engine/grid.py
```diff
@@ -435,26 +524,24 @@ def get_border_lawnmower_waypoints(self):
         """
         waypoints = []
         node_list = self.nodes
```
AC408, line 526: does node_list only contain the active nodes (nodes that we will traverse in lawnmower) or all nodes (including the blue inactive nodes)?

AC408 (changes requested):
I left some comments

engine/grid.py
jinlin270: Ah sorry the first if branch is not necessary

engine/grid.py
jinlin270: That makes sense, I fixed it

engine/grid.py
jinlin270: Can you also checks if it appends the last point at the center of the node? The last curve looks like it ended a little off the center.
Screen Shot 2022-11-14 at 12 08 41 AM

engine/grid.py
AC408: done

engine/grid.py
AC408: done

engine/grid.py
AC408: done

engine/grid.py
AC408: i dont see anywhere it is used

engine/grid.py
AC408: so remove both?

engine/grid.py
AC408: ok, resolve when finished

engine/grid.py
AC408: ok, can you add a comment?

engine/grid.py
AC408: nope; although it has been tested when adding waypoints

engine/grid.py
jinlin270: done

engine/grid.py
jinlin270: it looks like it is the entire grid. It is defined like this
`    self.nodes = generate_nodes(
            lat_min, long_min, self.num_rows, self.num_cols, STEP_SIZE_METERS)`
`generate_nodes Returns a 2D Numpy array of Node objects that make up the entire grid. Has dimensions [rows x cols]`

AC408: do you know why we only check the 0th col and the self.num_rows-1 col for get_border_lawnmower_waypoints?

AdityaKompella6 (approved):
I think this good but like do is the readius of curvature for the traversal big enough like the diameter of the circle for the sock?

sazhang02:
whats the plan for this pr?

PR #106, opened 2022-11-19
Allow adjustable grid selection of the entire gridspace

sazhang02:
### Description

Changed is_border_node check to make sure that there isn't a duplicate column.
Before fix:
Screen Shot 2022-11-19 at 12 35 37 AM
After fix:
Screen Shot 2022-11-19 at 12 35 51 AM

Also, changed the bounds of the positions we look at in get_all_lawnmower_waypoints_adjustable so we don't go out of bounds
Screen Shot 2022-11-20 at 12 03 50 AM

This change is reflected in the new circle.png as well and the screenshot has been updated such that all nodes on the border (and none extra) are marked as border nodes: 
Screen Shot 2022-11-20 at 12 41 37 AM

## Test Plan

python -m tests.functionality_tests.test_activation_shapes

Ran previous tests to check they are still working + matches previous test screenshots.

Checked current circle traversal script doesn't have double border nodes at the top.
Screen Shot 2022-11-20 at 12 12 19 AM

Change script to use a full triangle by each side of the activation, one at a time.
triangle_full_right
triangle_full_bottom
triangle_full_top
triangle_full_left

Change script to use a full circle by each side of the activation, one at a time.
circle_full_right
circle_full_bottom
circle_full_top
circle_full_left

Likewise, change script to use a full grid by uncommenting `grid.activate_rectangle(rec_row_start, rec_col_start, rec_row_limit, rec_col_limit)`
Screen Shot 2022-11-20 at 12 06 32 AM

These images for the full grid have been added to the folder `full_grid`.

## Other
Would like to merge this into lawnmower, not main


engine/grid.py
```diff
@@ -393,6 +399,8 @@ class Direction(Enum):
         while (phase != WaypointPhase.TERMINATE):
             if direction == Direction.LEFT:
                 new_pos = (curr_pos[0]-1,curr_pos[1])
+                if not self.is_position_in_grid_bounds(new_pos):
```
jinlin270, line 402: currently, this break statement seems to terminate the traversal prematurely, so the entire grid is not traversed. I tested with a rectangle traversal with ` grid.activate_rectangle(0, 0, rows, cols)`

engine/grid.py
```diff
@@ -407,6 +415,8 @@ class Direction(Enum):
                         phase = WaypointPhase.TERMINATE
             if direction == Direction.RIGHT:
                 new_pos = (curr_pos[0]+1,curr_pos[1])
+                if not self.is_position_in_grid_bounds(new_pos):
```
jinlin270, line 418: Same here, terminate the traversal prematurely

jinlin270 (changes requested):

engine/grid.py
sazhang02: oh yes! that makes sense I totally missed the purple line o7

sazhang02:
made changes + updated tests (w/ more tests!)

jinlin270:
no longer needed

PR #108, opened 2023-02-04
New sock traversal

jinlin270:
## Summary
New sock traversal is updated with optional vertical traversal.

### Description

New sock traversal is updated with optional vertical traversal which can be activated by inputting True when prompted "Vertical traversal:" when running test_activation_shapes.py. 

test_activation_shapes.py performs a functionality test for a case-by-case scenario for each shape. In this script, we have the eng quad made with a fixed lat long. Our grid code represents the eng quad with 34 rows 38 cols (which is why there shouldn’t be any user input for this btw).

Then the script asks the user which of the 3 shapes we are checking - rectangle, triangle, circle. For each shape, this script has an exact size hardcoded for what it should be (E.g. rectangle will take 1/3 of the grid, circle is set at the center w a radius of 2/3's of the total grid size).

This script will take in the shape input and whether vertical is "True" and show you a matplotlib graph of the guided traversal under these conditions. To check if the code is working you can go to the activation shapes folder and check if the graph that pops up is the same as the picture in the folder. This means the code still works the same for this case scenario (this is how this is a test). In the future, we may want to automate checking the output of the scripts to the expected images in the folder to make sure our grid code still works holistically

## Test Plan

The test plan is to test the rectangle, triangle, and circle with is_vertical = True/ False in test_activation_shapes.py and check the outputted traversal graph with the graphs in tests/functionality_tests/activate_shape_graphs. 
Note: The Test with shape = Triangle and Vertical = True is not expected to work due to its slanted orientation, but everything else should work.

## Other
Future TODOs :

Rename existing grid traversal types: lawnmower and lawnmower_b https://github.com/cornellnexus/src/pull/108/files#r1126797443
Change function parameters from individual x,y to a point (x,y) and update specifications to define this parameter and precondition that it's a tuple of ints https://github.com/cornellnexus/src/pull/108/files#r1130206826.
Change implementation for get_all_guided_lawnmower_waypoints_adjustable(self, is_vertical) to work for triangle/slanted shapes https://github.com/cornellnexus/src/pull/108/files#r1129259292
Improvements https://github.com/cornellnexus/src/pull/108#discussion_r1134750910
Separate out each shape into individual scripts
Create a interactive visualization to allow users to play around and determine what they see
[Future] Look into render testing system to automatic image comparison testing
Grid bug https://github.com/cornellnexus/src/pull/108#discussion_r1124926224


tests/functionality_tests/test_activation_shapes.py
```diff
@@ -44,11 +71,22 @@ def main():
         y3 = cols // 6 ## Good value 5
         grid.activate_triangle(x1,y1,x2,y2,x3,y3) 
         grid.find_border_nodes()
-        way_points = grid.get_all_lawnmower_waypoints_adjustable()
+        way_points = grid.get_all_guided_lawnmower_waypoints_adjustable(is_vertical)
 
+    if activation_type == "line":
+        row = 0 # rows are y position
+        col = cols // 2 # cols are x position
+        grid.activate_line(row, col, n=15, isHorizontal=False)
```
AC408, line 79: is n=15 something that works for all lines?

AC408 (approved):
Everything looks good. Just make sure to resolve the merge conflicts

tests/functionality_tests/test_activation_shapes.py
jinlin270: I think this is hardcoded for a specific case of a vertical line going straight up by 15.

sazhang02 (changes requested):
Please run the script after merging main!
`python -m tests.functionality_tests.test_activation_shapes`

There is also a bug now when running these script but I'm not sure where it's stemming from
Screen Shot 2023-03-03 at 1 48 26 PM

It might be related to this the issue I investigated before in this PR that was manually added to this branch? Not sure if something was changed when main was merged https://github.com/cornellnexus/src/pull/106/files


Also, is there a reason for why vertical traversal is missing the second to last node? Since horizontal is different
Screen Shot 2023-03-03 at 1 48 00 PM

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -44,11 +71,22 @@ def main():
         y3 = cols // 6 ## Good value 5
         grid.activate_triangle(x1,y1,x2,y2,x3,y3)
```
Shreya1233, line 72: When the activation type is triangle, an error is thrown for isInsideTriangle since it's not initialized in the Grid class. 
 if self.isInsideTriangle(x1, y1, x2, y2, x3, y3, x, y):
AttributeError: 'Grid' object has no attribute 'isInsideTriangle'

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -25,15 +52,15 @@ def main():
...
         grid.activate_rectangle(rec_row_start, rec_col_start, rec_row_limit, rec_col_limit)
         grid.find_border_nodes()
-        way_points = grid.get_all_lawnmower_waypoints_adjustable()
+        way_points = grid.get_all_guided_lawnmower_waypoints_adjustable(is_vertical)
 
     if activation_type== "circle":
         circle_center_row = rows // 2 ## Good value 19
         circle_center_col = cols // 2 ## Good value 17
         circle_radius = min(rows, cols) // 2 ## Good value 17
         grid.activate_circle(circle_center_row, circle_center_col, circle_radius)
```
Shreya1233, line 61: Sammie did leave a comment earlier saying that we should no longer ask users for row/col input and instead fix it to 34 rows, 38 cols. Right now, the users can set a row/col input (ex. 60,60)  so that the center becomes out-of-bounds. 

Screen Shot 2023-03-03 at 4 50 23 PM

Although cases like this will be resolved once the row/cols are fixed, there's an edge case where the center could be 30,30 (in-bounds) and the radius is 20. Should this case be allowed? If not, then we need to account for it here.

Shreya1233 (changes requested):
I noticed the same issues that Sammie mentioned in her comments (code compilation issue, 2nd to last node excluded in vertical traversal, the last two cols of the grid being red/identified as border nodes) but also added a few about the triangle activation case and the circle activation case.

tests/functionality_tests/test_activation_shapes.py
AC408: ^ any ideas?

engine/mission.py
```diff
@@ -14,7 +14,7 @@ class ControlMode(Enum):
     """
     LAWNMOWER = 1
     LAWNMOWER_B = 2
```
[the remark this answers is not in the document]
sazhang02: good point lets do lawnmower and lawnmower_b in a separate small pr

engine/grid.py
```diff
@@ -301,79 +320,379 @@ def find_border_nodes(self):
...
+            self.direction = self.direction.UP
+        while not self.waypoints_is_finished:
+            if self.curr_pos == None:
+                self.waypoints_is_finished = True
+
+            # Question: can we delete? bounds checked in add guided? change to assert?
+            # elif (self.curr_pos[0] > self.num_rows - 1 or 
+            #     self.curr_pos[0] < 0 or self.curr_pos[1] > self.num_cols - 1 or 
+            #     self.curr_pos[1] < 0):
+            #     self.phase = self.WaypointPhase.TERMINATE
```
sazhang02, line 686:
```suggestion
```
Do uk if we still need this? If so, let's mention it in pr summary/other. If not, suggested deletion.

tests/functionality_tests/test_activation_shapes.py
sazhang02: ^^ maybe n = cols//2 for half the width of the grid? agree we should code it dynamically

engine/grid.py
```diff
@@ -301,79 +320,379 @@ def find_border_nodes(self):
...
+            return
+
+        # Column where we want to begin turning, one node closer to the inside of our shape
+        turning_column = self.get_turning_column(edge_column) 
+        next_row = self.curr_pos[1] + 1
+
+        # Question: Can we delete?
+        # elif curr_pos[0] < (left_pos[0] + 1):
+        #     phase = WaypointPhase.TERMINATE
+        # else:
```
sazhang02, line 581:
```suggestion
```
I feel like we can delete this?

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -25,15 +52,15 @@ def main():
...
         circle_center_row = rows // 2 ## Good value 19
         circle_center_col = cols // 2 ## Good value 17
         circle_radius = min(rows, cols) // 2 ## Good value 17
         grid.activate_circle(circle_center_row, circle_center_col, circle_radius)  
         grid.find_border_nodes()
-        way_points = grid.get_all_lawnmower_waypoints_adjustable()
+        way_points = grid.get_all_guided_lawnmower_waypoints_adjustable(is_vertical)
 
     if activation_type== "triangle":
         x1 = rows // 6## Good value 6
```
sazhang02, line 66:
```suggestion
    # TODO: Vertical traversal in triangle
    if activation_type== "triangle":
        x1 = rows // 6## Good value 6
```
Let's make a note that triangle vertical traversal doesn't work

engine/grid.py
```diff
@@ -301,79 +320,379 @@ def find_border_nodes(self):
...
+            and turning counter clock wise at the end of the row (or terminating).
+        The plot_circle plots the circular waypoints during the turn
+        """
+        # TODO: move find border nodes into this func instead of doing it
+        # everytime before this is called
+        self.curr_pos = self.leftmost_node_pos
+        self.waypoints.append(self.curr_pos)
+        if is_vertical:
+            self.direction = self.direction.UP
+        while not self.waypoints_is_finished:
```
sazhang02, line 678:
```suggestion
        if is_vertical:
            # TODO: Implement vertical traversal for triangle shape
            self.direction = self.direction.UP
        while not self.waypoints_is_finished:
```
Let's make a note that triangle vertical traversal doesn't work

sazhang02 (changes requested):
Added some more comments based on the edge of grid fix + noting current triangle vertical bug. 

Re-request + ping me on slack when all previous PR comments are addressed/resolved.

engine/grid.py
```diff
@@ -202,40 +222,32 @@ def get_neighbor_node(self, row, col, row_max, col_max):
         else:
             return neighbor_node
 
-
+    ##Activates rectangle based on row start/end, col start/end
```
AC408, line 225:
```suggestion
```

engine/grid.py
```diff
@@ -202,40 +222,32 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
         """
         for x in range(row, row_limit):
             for y in range(col, col_limit):
-                self.nodes[x,y].is_active = True
+                self.nodes[x, y].is_active = True
         return self.nodes
 
-
-    
+    ##Activates circle based on center and radius
```
AC408, line 236:
```suggestion
```

engine/grid.py
```diff
@@ -202,40 +222,32 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
                     self.nodes[x, y].is_active = True
 
-   
-
-    def is_inside_triangle(self, x1, y1, x2, y2, x3, y3, x, y):
-        """
-        A function to check whether point P(x, y) lies inside the triangle 
-        formed by A(x1, y1), B(x2, y2) and C(x3, y3)
-        """
+    # A function to check whether point P(x, y)
```
AC408, line 246: I think I actually prefer this as a spec than a comment. Should also specify the function param (description, type), and return (description, type)

engine/grid.py
```diff
@@ -253,44 +265,51 @@ def area(x1, y1, x2, y2, x3, y3):
...
+                self.nodes[col+i][row].is_active = True
         else:
-            return False
+            for i in range(n):
+                self.nodes[col][row+i].is_active = True
 
+    ##Activate triangle based on three points
     def activate_triangle(self, x1, y1, x2, y2, x3, y3):
         """
-        Activates all the nodes in a traingle based on three points.
```
AC408, line 263: why change this spec? I think it would be nice to be specific. If the problem is redundancy (x color of 1 point, x color of 2nd point), can do currying and uncurrying -> param takes in 3 points instead of 3 x coor and 3 y coor. Can then specify that the points are vertices of the triangle

engine/grid.py
```diff
@@ -253,44 +265,51 @@ def area(x1, y1, x2, y2, x3, y3):
...
+        for x in range(self.num_rows):
+            for y in range(self.num_cols):
+                if self.isInsideTriangle(x1, y1, x2, y2, x3, y3, x, y):
                     self.nodes[x, y].is_active = True
 
-    
+    ##Checks if node is on border of activated nodes
     def is_on_border(self, row, col, row_limit, col_limit):
-        """
-        Checks if node is on border of activated nodes.
```
AC408, line 275: I think spec would be nice. I'm assuming row and col is coor of node? Might make more sense to curry it into point and in the body of the code do (row, col) = point. Not sure about syntax of prev sentence, but you get the point

engine/grid.py
```diff
@@ -253,44 +265,51 @@ def area(x1, y1, x2, y2, x3, y3):
...
-        max_col = min(col_limit, col+1)
-        max_row = min(row_limit, row+1)
+        min_col = max(0, col - 1)
+        min_row = max(0, row - 1)
+        max_col = min(col_limit, col + 1)
+        max_row = min(row_limit, row + 1)
 
         # If this node is on the very edge of the grid, it is automatically a border node
-        if min_col == 0 or min_row == 0 or max_col == col_limit or max_row == row_limit:
+        if row == 0 or col == 0 or col == self.get_num_cols() - 1 or row == self.get_num_rows() - 1:
```
AC408, line 302: nit, could change self.get_num_cols() -1 into var called last_col or smthg to be more descriptive about what this condition means

engine/grid.py
```diff
@@ -301,80 +324,371 @@ def find_border_nodes(self):
...
+                        self.nodes[row][col].is_border = True
+                        border_list.append((node, row, col))
+                        if leftmost_node_pos is None or row < leftmost_node_pos[0]:
+                            leftmost_node = node
+                            leftmost_node_pos = (row, col)
+
+            self.border_nodes = border_list
+            self.leftmost_node = leftmost_node
+            self.leftmost_node_pos = leftmost_node_pos
+
```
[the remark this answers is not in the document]
AC408: I think some renaming/spec would be nice. Leftmost_node can indicate leftmost inactive node. Also, the function also fills out the border_nodes, which could be done in another function or the fun name should change.

engine/grid.py
```diff
@@ -301,79 +320,379 @@ def find_border_nodes(self):
...
+        If there are no more nodes in the next row, return None.
+
+        pos: the current position (col,row)
+        """
+        # Obtain all border nodes in the next and current rows 
+        if self.direction == self.Direction.RIGHT or self.direction == self.Direction.LEFT:
+            # Note: node_info: (node, row, col)
+            #Obtain all border_nodes on the next col
+            
+            candidate_nodes_next = [node_info for node_info in self.border_nodes if node_info[2] == pos[1] + 1]
```
AC408, line 398: maybe a comment about what candidate_nodes_next is? same with candidate_nodes_curr

engine/grid.py
```diff
@@ -301,79 +320,379 @@ def find_border_nodes(self):
...
+            candidate_nodes_curr = [node_info for node_info in self.border_nodes if node_info[1] == pos[0]]
+        
+        if (candidate_nodes_next == [] and candidate_nodes_curr == []):
+            # There are no border nodes in the current row or next row
             return None
+        elif (candidate_nodes_next == []):
+            # Since there we have reached the last row that we can traverse, 
+            # let's traverse this last row
+            if self.direction == self.Direction.RIGHT:
+                node_info_curr = max(candidate_nodes_curr, key=lambda node_info: node_info[1])
```
AC408, line 413: what is node_info_curr?

engine/grid.py
```diff
@@ -301,79 +320,379 @@ def find_border_nodes(self):
...
+                return max(node_info_next[1], node_info_curr[1])
+            elif self.direction == self.Direction.UP:
+                node_info_next = max(candidate_nodes_next, key=lambda node_info: node_info[2])
+                node_info_curr = max(candidate_nodes_curr, key=lambda node_info: node_info[2])
+                return min(node_info_next[2], node_info_curr[2])
+            elif self.direction == self.Direction.DOWN:
+                node_info_next = min(candidate_nodes_next, key=lambda node_info: node_info[2])
+                node_info_curr = min(candidate_nodes_curr, key=lambda node_info: node_info[2])
+                return max(node_info_next[2], node_info_curr[2])
+
```
AC408, line 442: duplicate code with above code

engine/grid.py
```diff
@@ -301,79 +320,379 @@ def find_border_nodes(self):
...
+        """
+        Returns a circle of nodes starting from the [start_pos], going at orientation [orientation], and ending at the
+        [end_pos] with center [center].
+
+        Arguments:
+            start_pos: float tuple representing the starting point of the circle being plotted
+            end_pos: float tuple representing the ending point of the circle being plotted
+            center: float tuple representing the center point of the circle being plotted
+            orientation: orientation that the nodes are being plotted from, starting with the start_pos and ending
+                at the end_pos
```
AC408, line 453: lmao, what happened here. why is it on another line

AC408 (changes requested):
Some comments, esp on grid (whoops, didn't realize that was changed. had to click view diff to see changes)

engine/grid.py
```diff
@@ -301,79 +320,379 @@ def find_border_nodes(self):
...
+                return min(node_info_next[2], node_info_curr[2])
+            elif self.direction == self.Direction.DOWN:
+                node_info_next = min(candidate_nodes_next, key=lambda node_info: node_info[2])
+                node_info_curr = min(candidate_nodes_curr, key=lambda node_info: node_info[2])
+                return max(node_info_next[2], node_info_curr[2])
+
+    def plot_circle(self, start_pos, end_pos, center, orientation, theta_step=math.pi / 12):
+        """
+        Returns a circle of nodes starting from the [start_pos], going at orientation [orientation], and ending at the
+        [end_pos] with center [center].
```
Shreya1233, line 446: fix indentation here

Shreya1233 (commented):
I think the main bugs are resolved with the new commits. The only main things I have to comment are about the circle edge case + I noticed that the 2nd last node for vertical traversal is still missing and if that was intentional?

tests/functionality_tests/test_activation_shapes.py
jinlin270: Changed, the lines are vertical going up, but our rows/columns are swapped, so it will be cols//2  for half the height of the grid

engine/grid.py
```diff
@@ -253,44 +269,51 @@ def area(x1, y1, x2, y2, x3, y3):
...
+        # Note: rows are y-position and columns are x-position
+        if isHorizontal:
+            for i in range(n):
+                self.nodes[col+i][row].is_active = True
         else:
-            return False
+            for i in range(n):
+                self.nodes[col][row+i].is_active = True
 
+    ##Activate triangle based on three points
```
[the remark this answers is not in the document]
jinlin270: done

engine/grid.py
```diff
@@ -81,7 +95,7 @@ def generate_nodes(start_lat, start_long, rows, cols, step_size_m):
             # traversal.
             for i in range(cols):
                 for j in range(rows):
-                    is_border = (j == 0 or j == rows - 1)
+                    # is_border = (j == 0 or j == rows - 1)
```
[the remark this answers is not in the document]
jinlin270: That is interesting because my code currently works with this line commented out, and there are no references to this variable in my file.

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -7,16 +7,43 @@ def main():
...
+    try: 
+        rows = int(input("number of rows:"))
+    except:
+        rows = grid.get_num_rows()
+
+
+    try: 
+        cols = int(input("number of cols:"))
+    except:
+        cols = grid.get_num_cols()
```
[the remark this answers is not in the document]
jinlin270: Does this mean we should remove the input option?

engine/grid.py
```diff
@@ -301,80 +324,371 @@ def find_border_nodes(self):
...
+        else:
+            # Turn to next row
+            # Create waypoints needed for a smooth turning trajectory to 
+            # "guide" our robot to the next original waypoint
+            circle_plt = self.plot_circle((self.curr_pos[0], turning_row), (next_column, turning_row),
+                                        (self.curr_pos[0] + .5, turning_row), self.get_turn_orientation())                                       
+            self.waypoints += circle_plt
+            # Update traversal details
+            self.switch_directions()
+            # self.curr_pos = (turning_column, next_row)
```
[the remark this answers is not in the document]
jinlin270: I will clean out all the commented out code

tests/functionality_tests/test_activation_shapes.py
AdityaKompella6: I think it should be allowed but I am just wondering what the red nodes are. If they are border nodes shouldn't that be incorrect? Maybe when Andrew fixes the border function it will work.

engine/grid.py
```diff
@@ -301,80 +324,371 @@ def find_border_nodes(self):
...
+            self.curr_pos = (next_column, turning_row)
```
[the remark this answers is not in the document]
jinlin270: Done!

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -44,11 +71,22 @@ def main():
...
+        grid.find_border_nodes()
+        way_points = grid.get_all_guided_lawnmower_waypoints_adjustable(is_vertical)
 
-    for i in range(rows):
-        for j in range(cols):
+    # Questions: what if we make a new file for visualizing the path? It would also help with catching exceptions.
+    #  For example, `node = grid.nodes[i,j]` can throw index out of bound error given STEP_SIZE_METERS of 100,
+    #   num rows and cols 3 and circle activation type because num_y_step and num_x_step in int and inversely
+    #   proportional to STEP_SIZE_METERS, so num_y_step and num_x_step is 0. Same problem when inputting row and col.
+    #   Not sure the formula for determining the max row and col to input tho
```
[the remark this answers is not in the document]
jinlin270: It look like Allen added the question 4 months ago, but it may still be relevant? I will add mention it in the PR

tests/functionality_tests/test_activation_shapes.py
jinlin270: Thank you Shreya, this is very interesting. We should definitely discuss this during the meeting. If we do want to consider cases this this, we would need to fix the initialization.

AdityaKompella6 (approved):
Good job overall the sock works well. After border node fixes it will probably work better

engine/mission.py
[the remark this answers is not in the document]
jinlin270: good point. I am not sure what they stand for

engine/grid.py
jinlin270: I don't think we need this, I deleted them

tests/functionality_tests/test_activation_shapes.py
jinlin270: Yes, but we did talk about this last meeting. The vertical traversal currently does not work because of the weird orientation. But we could make a separate triangle test for vertical traversal.

engine/grid.py
jinlin270: I don't think we should change the traversal implementation, just the triangle test for the vertical traversal.

engine/grid.py
jinlin270: Cleaned up

engine/grid.py
jinlin270: Deleted, thank you!

engine/grid.py
jinlin270: Screen Shot 2023-03-09 at 4 26 29 PM

engine/grid.py
jinlin270: This is a good point, I will change the inputs to points

engine/grid.py
jinlin270: good points. Also there are a bunch of other functions which also have row and col, so we can prob curry that into points as well

tests/functionality_tests/test_activation_shapes.py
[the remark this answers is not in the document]
AC408: I remember somewhat the example exception in this comment, but I'm don't quite remember what I meant by visualizing the path 😅

engine/grid.py
jinlin270: It is the last node we would traverse in the grid. Comments added

engine/grid.py
jinlin270: refactored

engine/grid.py
jinlin270: 🤷

engine/grid.py
```diff
@@ -202,178 +223,435 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
         else:
-            return False
+            for i in range(n):
+                self.nodes[col][row+i].is_active = True
 
-    def activate_triangle(self, x1, y1, x2, y2, x3, y3):
+    def activate_triangle(self, p1, p2, p3):
         """
-        Activates all the nodes in a traingle based on three points.
+        Activates all the nodes in a traingle based on three points p1, p2, p3.
```
AC408, line 284:
```suggestion
        Activates all the nodes in a traingle based on three vertices p1, p2, p3.
```

engine/grid.py
```diff
@@ -202,178 +223,435 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
+            else:
+                circle_plt = self.plot_circle((turning_column, self.curr_pos[1]), (turning_column, next_row),
+                                              (turning_column, self.curr_pos[1] + .5), self.get_turn_orientation())
+            self.waypoints += circle_plt
+            # Update traversal details
+            self.switch_directions()
+            if is_vertical:
+                self.curr_pos = (next_row, turning_column)
+            else:
+                self.curr_pos = (turning_column, next_row)
```
AC408, line 624: assuming curr_pos is (row, col), might be mislead to leave var names as turning_column for row arg and next_row for col arg

engine/grid.py
```diff
@@ -118,11 +134,17 @@ def generate_nodes(start_lat, start_long, rows, cols, step_size_m):
         self.active_waypoints_list = []
         self.inactive_waypoints_list = []
 
+        # Properties used to determine our adjustable grid traversal pathing
+        self.waypoints_is_finished = False
+        self.curr_pos = None
```
AC408, line 139: might want to specify curr_pos is (row, col) if vertical and (col, row) if not vertical

AC408 (changes requested):
small nitpicking comments, but otherwise looks good. I can't approve without all the tests succeeding though. Can you check and resolve that issue?

engine/grid.py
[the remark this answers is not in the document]
sazhang02: fixed!

AC408 (approved):

tests/functionality_tests/test_activation_shapes.py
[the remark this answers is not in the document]
sazhang02: oop LOL

tests/compilation_tests/grid_test.py
```diff
@@ -109,19 +109,19 @@ def test_node_boundaries(self):
...
-        self.assertFalse(g.is_inside_triangle(2, 2, 3, 3, 1, 3,
-                         10, 10), "this point is outside the triangle")
-        self.assertTrue(g.is_inside_triangle(2, 2, 3, 3, 1, 3, 2,
-                        2), "this point is inside the triangle")
+        self.assertTrue(g.isInsideTriangle((2, 2), (3, 3), (1, 3), (2,
+                        2.5)), "this point is inside the triangle")
+        self.assertFalse(g.isInsideTriangle((2, 2), (3, 3), (1, 3),
+                         (10, 10)), "this point is outside the triangle")
+        self.assertTrue(g.isInsideTriangle((2, 2), (3, 3), (1, 3), (2,
+                        2)), "this point is inside the triangle")
```
sazhang02, line 117: should be snake case not camel case

engine/grid.py
sazhang02: wdym by vertical? if direction is set to vertical?

engine/grid.py
```diff
@@ -202,178 +222,435 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
-    def is_inside_triangle(self, x1, y1, x2, y2, x3, y3, x, y):
-        """
-        A function to check whether point P(x, y) lies inside the triangle 
-        formed by A(x1, y1), B(x2, y2) and C(x3, y3)
-        """
-        def area(x1, y1, x2, y2, x3, y3):
-
-            return abs((x1 * (y2 - y3) + x2 * (y3 - y1)
-                        + x3 * (y1 - y2)) / 2.0)
+    def isInsideTriangle(self, p1, p2, p3, p4):
```
sazhang02, line 244: change name throughout code and when function is used to be snake case

engine/grid.py
```diff
@@ -202,178 +222,435 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
+        """ A function to check whether point P(x, y) lies inside the triangle formed by
+        A(x1, y1), B(x2, y2) and C(x3, y3)"""
```
sazhang02, line 246: If we are changing to p1, p2, p3, p4 we should specify the parameters and this precondition in the specification (e.g. p1,p2,... are tuples representing (x,y) points on our grid

engine/grid.py
```diff
@@ -202,178 +222,435 @@ def get_neighbor_node(self, row, col, row_max, col_max):
...
+        return A == (A1 + A2 + A3)
+
+    def activate_line(self, row, col, n, isHorizontal):
+        """
+        Activate a single horizontal line of length n starting at row, col.
+        If isHorizontal is True, the line is from (row, col) to (row, col + n).
+        Otherwise, the line is from (row, col) to (row + n, col).
+        """
+        # Note: rows are y-position and columns are x-position
+        if isHorizontal:
```
sazhang02, line 274: change isHorizontal to is_horizontal (snake case)

engine/grid.py
sazhang02: good point from allen, lets do this in a separate pr!!

engine/grid.py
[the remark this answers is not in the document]
sazhang02: ^^

engine/grid.py
sazhang02: hmm the all shape tests uses this code tho to create the path right?

The grid object's active nodes are set in one of the activate shape functions and then the path is calculated here regardless of shape? Feel like we should note somewhere that this function doesn't get the path for all waypoints (case of waypoints attribute is set w slanted or specific triangle orientations).

sazhang02:
Future TODOs (add this to the Other section of PR summary @jinlin270 and feel free to edit!): 
- Rename existing grid traversal types: lawnmower and lawnmower_b https://github.com/cornellnexus/src/pull/108/files#r1126797443
- Change function parameters from individual x,y to a point (x,y) and update specifications to define this parameter and precondition that it's a tuple of ints https://github.com/cornellnexus/src/pull/108/files#r1130206826.
- Change implementation for `get_all_guided_lawnmower_waypoints_adjustable(self, is_vertical)` to work for triangle/slanted shapes https://github.com/cornellnexus/src/pull/108/files#r1129259292

tests/functionality_tests/test_activation_shapes.py
sazhang02: Oh, then let's either mention that its for the specific orientations you found or change this test script to use a case that works (bc it could be confusing to someone why triangle doesn't have a line in it when we run the script)

sazhang02: actually both would prob be best

tests/functionality_tests/test_activation_shapes.py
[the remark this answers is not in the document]
sazhang02: yes!

Explanation/reasoning copied from slack for documentation purposes

If we want to keep this as a functionality test we should delete it.

Just for some more context, functionality tests should be scripts testing a specific scenario. In this case the original motivation was to turn Aditya’s interactive script for the activation shapes into a script testing a specific case scenario — in this case the space determine by the lat/long of the eng quad (which converts to 39 columns, 34 rows using our grid code). In an idea case we would just have to run the script and compare it with the expected output/image to make sure it’s working. This idea can be confusing but a good example of this is Allen’s obstacle avoidance script obstacle_tracking_test.py, which mocks a scenario written in csv files, runs the code under that case, and checks if the output is as expected. By having a static cases, we can quickly check even without much understanding + potentially expand to using render tests to automatically check image-based functionality tests (like we do rn on github for the compilation tests).

In the case where we want to determine the bounds rows/cols of the grid, that would be more of a visualizer (if we preset inputs then grid pops up) or a interactive gui/visualizer (if we pop up w grid and allow dynamic changes like changing shape).

Though it is good for debugging and playing around w the code, I would say the try except to determine the number of rows/cols isn’t the clearest interaction for a visualizer since there are a lot of ambiguity (e.g. would need to look at code to know pressing “enter” leads to a default of the entire grid bounds being used, need to tell users to enter valid numbers, etc). Additionally,

```
plt.xlim(-1, grid.get_num_rows()+1)
plt.ylim(-1, grid.get_num_cols()+1)
```

shows the size 39, 34 regardless (which is why the circle case Shreya brought up becomes off screen.
I think it would be good to create a visualizer later on since it’ll require more changes than keeping this as a functionality script (easy future onboarding task o7). Now that I’m thinking about this again, to be more of a functionality test, it shouldn’t have any user input even for the shapes and instead be in separate files or functions or something — but lets make that a later issues xD

I think the best approach would be to
- Keep this as a functionality test, testing the adjustable grid code on the eng quad
- Separate out each shape into individual scripts
- Create a interactive visualization to allow users to play around and determine what they see

tests/functionality_tests/test_activation_shapes.py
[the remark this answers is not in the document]
sazhang02: since we don't really remember why i feel like we should delete it??

tests/functionality_tests/test_activation_shapes.py
sazhang02: The circle case is bc we are allowing user inputs. Should be resolved when we remove input space since we are fixing the matplotlib to the eng quad size
Explained here:
https://github.com/cornellnexus/src/pull/108#discussion_r1134750910

sazhang02 (changes requested):
Looking good! let's finish changes for unresolved comments like https://github.com/cornellnexus/src/pull/108/files#r1124906029 and would like to address https://github.com/cornellnexus/src/pull/108#pullrequestreview-1324420700 as mentioned before (add image snapshots + pr summary)

sazhang02:
Future TODOs (add this to the Other section of PR summary @jinlin270 and feel free to edit!): 
- Rename existing grid traversal types: lawnmower and lawnmower_b https://github.com/cornellnexus/src/pull/108/files#r1126797443
- Change function parameters from individual x,y to a point (x,y) and update specifications to define this parameter and precondition that it's a tuple of ints https://github.com/cornellnexus/src/pull/108/files#r1130206826.
- Change implementation for `get_all_guided_lawnmower_waypoints_adjustable(self, is_vertical)` to work for triangle/slanted shapes https://github.com/cornellnexus/src/pull/108/files#r1129259292
- Improvements https://github.com/cornellnexus/src/pull/108#discussion_r1134750910
  - Separate out each shape into individual scripts
  - Create a interactive visualization to allow users to play around and determine what they see
  - [Future] Look into render testing system to automatic image comparison testing

tests/functionality_tests/test_activation_shapes.py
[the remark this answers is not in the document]
AC408: I'm fine with deleting making new file for visualizing path. But I think the example should be kept if the problem still exists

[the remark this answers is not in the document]
AC408: Yeah, I just tested 3 rows, 3 cols, circle, True, and set STEP_SIZE_METERS in grid to 100 and the test crashes. If we assume we aren't going to have the STEP_SIZE_METERS change, then I think it'll be fine to just leave that comment in but copy it to the grid file as well.

engine/grid.py
```diff
@@ -81,21 +97,20 @@ def generate_nodes(start_lat, start_long, rows, cols, step_size_m):
             # traversal.
             for i in range(cols):
                 for j in range(rows):
-                    is_border = (j == 0 or j == rows - 1)
```
AC408, line 84: I feel like I remember another comment about test case failing because is_border isn't initialize. I'm not sure if the new code is pushed but when I tested it, first of all, I couldn't pull because there was a bunch of merge conflicts, but if I don't pull and run the test, it crashed and said this wasn't initialized

AC408: If the code was pushed along with the merging main step, make sure to keep commits as separate as possible

sazhang02: I made the change u mentioned in commit https://github.com/cornellnexus/src/pull/108/commits/4137dcf81c4ab1d798c42a64afc67c962176c27b bc we defined what a border node was differently before so i moved the is_border node definition to get_all_lawnmower

sazhang02: a bit confused about what wasn't working?

engine/grid.py
AC408: yeah, based on is_vertical

engine/grid.py
AC408: Merge conflict when pulling

engine/grid.py
jinlin270: added

tests/compilation_tests/grid_test.py
jinlin270: thanks Sammy, changed

engine/grid.py
jinlin270: changed

engine/grid.py
jinlin270: I am not sure, self.get_num_cols() feels more descriptive the function does number of columns.  It is also used in other functions as well for that purpose, so I think it might be best to keep it.

engine/grid.py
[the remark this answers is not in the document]
sazhang02: nice!

engine/grid.py
sazhang02: agreed w lin lets keep it

sazhang02 (approved):
AMAZING JOB LIN!!! THANK U FOR ADDRESSING ALL CHANGES <3

PR #112, opened 2023-02-24
Abstracting attributes/data of robot class to robot_state class

sazhang02:
## References

[Using kwargs for initialization ](https://www.geeksforgeeks.org/args-kwargs-python/)
### Description

We have a bunch of attributes in our `Robot` class. We also have an issue with interrupting our ongoing Mission cleanly if we receive any transmission from the raspberry pi. In order to fix these issues, we can abstract all the attributes in our `Robot` class to another class called `Robot_State`. This allows us to keep all information about our robot separate (in case we need to interrupt and change these attributes in the future). A large majority of these changes are for the new variable access (e.g. `robot.robot_state.phase` instead of `robot.phase`.

During the refactoring, sensor dependency flags were changed to use `is_raspberrypi()` instead of passing down a `is_sim` flag or checking if True and manually changing the variable (esp in the electrical files). Some attributes were also renamed (e.g. `ekf_var` to `ekf`).

**Overall changes**
- Create `Robot_State` class and refactor attributes
- Update `sim_trajectory` to initialize Robot_State for Robot
- Rename attributes (e.g. `ekf_var` to `ekf`)
- Sensor dependency checks use `is_raspberrypi()`
- `Phase` abstracted to its own class
- Test cases that initialize `Robot` are updated to initialize `Robot_State` and pass that into `Robot`
- Duplicate test code for `traverse_standard_test`, `robot_test`, `execute_setup` consolidated into the robot_test folder (one test is commented out bc it wasn't passing + calculations seem off to me).

## Test Plan

Ran `python -m engine.sim_trajectory` to make sure sim_trajectory still works
Ran `python -m tests.compilation_tests.test_runner` to make sure all test cases depending on robot attributes were updated and tests still run

## Other

Co-authored by @Aayush-Agnihotri 

### TODOS
- Future `Mission` and `sim_trajectory` refactoring will build off of this refactoring work
- Re-evaluate attributes that were originally in `Robot` (and `moved` to Robot_State in this PR) + specifications. All comments with a 🚀 
- Discuss how to further refactor robot.py (https://github.com/cornellnexus/src/pull/112/files/8636700ac879a44d2623362472a239071c945908#r1130050539)
- Clean up random comments throughout codebase (e.g. https://github.com/cornellnexus/src/pull/112#discussion_r1130057426)
- Go through our test files (e.g. https://github.com/cornellnexus/src/pull/112#discussion_r1130061656)
- Manual control phase (e.g. https://github.com/cornellnexus/src/pull/112#discussion_r1131302097)
- Move non-unit test files from compilation_tests -> functionality_tests (e.g. movement_test and gps_test https://github.com/cornellnexus/src/pull/112#pullrequestreview-1333858307)
- Determine where to initialize our sensors (when we initialize robot vs execute_setup https://github.com/cornellnexus/src/pull/112#discussion_r1131327909)
- When we rereimplement pack transmission account for noise in the heading pid values (https://github.com/cornellnexus/src/pull/112#discussion_r1130018265)


engine/robot_state.py
```diff
@@ -0,0 +1,141 @@
...
+        if not self.is_sim:
+            self.motor_controller = MotorController(self, wheel_radius = 0, vm_load1 = 1, vm_load2 = 1, L = 0, R = 0)
+            self.robot_radio_session = RadioModule(serial.Serial('/dev/ttyS0', 57600)) 
+            self.gps = GPS(serial.Serial('/dev/ttyACM0', 19200, timeout=5)) 
+            self.imu = IMU(init_i2c = busio.I2C(board.SCL, board.SDA)) 
+        
+
+    # def rpiToGui():
+    #     pass
+
```
Aayush-Agnihotri, line 141: ```suggestion```

tests/compilation_tests/database_test.py
```diff
@@ -113,7 +120,7 @@ def test_get_data(self):
                     self.assertEqual(ans[1], data[1], name)
                     self.assertEqual(ans[2], data[2], name)
                 else:
-                    self.assertEqual(ans, database.get_data(name), name)
+                    self.assertEqual(ans, database.get_data(name), ans)
```
Aayush-Agnihotri, line 123: Why was this one changed to ans?

tests/compilation_tests/database_test.py
```diff
@@ -177,7 +184,8 @@ def test_update_data(self):
                     self.assertEqual(ans, database.get_data(name), name)
```
Aayush-Agnihotri, line 184: Should this one also be ans instead of name?

engine/robot.py
```diff
@@ -57,170 +26,76 @@ class Robot:
...
         # if it is a simulation,
-        if self.is_sim:
-            self.truthpose = np.append(
-                self.truthpose, np.transpose(self.state), 0)
+        if self.robot_state.is_sim:
+            self.robot_state.truthpose = np.append(
+                self.robot_state.truthpose, np.transpose(self.robot_state.state), 0)
+        else:
+            imu_data = self.robot_state.imu.get_imu()
+            self.robot_state.state[2] = math.degrees(math.atan2(imu_data["mag"][1], imu_data["mag"][0]))
```
Aayush-Agnihotri, line 86: What does the else branch of this if statement do - I'm not sure I understand

tests/compilation_tests/database_test.py
sazhang02: :0 good catch i changed it to debug

tests/compilation_tests/database_test.py
sazhang02: name o7

engine/robot.py
sazhang02: same 🤔 checked main and its not there?? think it might've been an incorrectly resolved merge conflict from main. removed!

sazhang02:
Tagging the whole team for transparency since this will change how we access data throughout our codebase!

Aayush-Agnihotri (approved):

engine/database.py
```diff
@@ -16,21 +16,21 @@ def __init__(self, robot):
             heading_pid: heading PID in format of [proportional factor, integral factor, derivative factor]
```
juliehelenek, line 16: I think this is just for simulating noise (I don't think it led to any fruitful results, just "our simulation doesn't work well if there is a lot of noise") - if that's the case, not sure if it should be a dedicated attribute

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+            plastic_weight: the weight of the trash the robot has collected
+            init_threshold (Double): Radius from initial position that will detect robot is back in initial position
+            goal_threshold (Double): Threshold from goal that will be detected as reaching goal in obstacle avoidance
+            noise_margin (Double): Margin from init_threshold that the robot has to leave before detecting robot has
+                left initial position
+        """
+        # User defined parameters
+        self.epsilon = epsilon
+        self.max_velocity = max_velocity
+        self.radius = radius
```
juliehelenek, line 71: Why do we need both radius and width? 
Also, the term radius is a little misleading because our robot is not a circle. Are we treating it as one?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+            move_dist: the distance in meters that the robot moves per time dt
+            turn_angle: the angle in radians that the robot turns per time dt regardless of time step
+            plastic_weight: the weight of the trash the robot has collected
+            init_threshold (Double): Radius from initial position that will detect robot is back in initial position
+            goal_threshold (Double): Threshold from goal that will be detected as reaching goal in obstacle avoidance
+            noise_margin (Double): Margin from init_threshold that the robot has to leave before detecting robot has
+                left initial position
+        """
+        # User defined parameters
+        self.epsilon = epsilon
```
juliehelenek, line 69: The epsilon attribute is interesting to me... in the comment it says it's defining how tight the turning is - are we assuming in every single situation that we are turning, it will turn the same amount? If not, isn't it better to make this attribute dependent on other values? (Maybe not because the sand will cause a lot of slipping, but just some food for thought?)

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.max_velocity = max_velocity
+        self.radius = radius
+
+        # Flags
+        self.is_sim = not is_raspberrypi()
+        self.is_store = False
+        self.phase = Phase.SETUP
+        self.avoid_obstacle = False
```
juliehelenek, line 77: Why do we need this flag? Can't we just look at the state?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+                left initial position
+        """
+        # User defined parameters
+        self.epsilon = epsilon
+        self.max_velocity = max_velocity
+        self.radius = radius
+
+        # Flags
+        self.is_sim = not is_raspberrypi()
+        self.is_store = False
```
juliehelenek, line 75: This attribute name kinda sucks, maybe something like should_store_data?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+    from electrical.motor_controller import MotorController
+    import electrical.gps as GPS 
+    import electrical.imu as IMU 
+    import electrical.radio_module as RadioModule
+    import serial
+    # from engine.sensor_module import SensorModule # IMU + GPS testing
+
+class Robot_State:
+    """
+    A class containing robot-specific information about the current state of a robot.
```
juliehelenek, line 17: Overall comment: The order of all of these attributes in the comment should match the order of all the attributes when defined

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        # Flags
+        self.is_sim = not is_raspberrypi()
+        self.is_store = False
+        self.phase = Phase.SETUP
+        self.avoid_obstacle = False
+        
+        # Constants
+        self.width = 700
+        self.time_step = 1
+        self.position_kp = 1
```
juliehelenek, line 82: i feel like we can make position_kp, position_ki, and position_kd into a 3-tuple instead of having 3 different attributes

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+            position_kp: the proportional factor of the position PID
+            position_ki: the integral factor of the position PID
+            position_kd: the derivative factor of the position PID
+            position_noise: the flat amount of noise added to the robot's phase on each localization step
+            heading_kp: the proportional factor of the heading PID
+            heading_ki: the integral factor of the heading PID
+            heading_kd: the derivative factor of the heading PID
+            heading_noise: ?
+            phase: the phase of the robot
+            time_step: the amount of time that passes between each feedback loop cycle, should only be used if is_sim
```
juliehelenek, line 58: why is this only used if is_sim? the time step is also involved in calculations for the control of the robot

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.avoid_obstacle = False
+        
+        # Constants
+        self.width = 700
+        self.time_step = 1
+        self.position_kp = 1
+        self.position_ki = 0
+        self.position_kd = 0
+        self.position_noise = 0
+        self.heading_kp = 1
```
juliehelenek, line 86: likewise, can probably put this in a 3-tuple

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+            is_sim: False if the physical robot is being used, True otherwise
+            is_store: False if csv data should not be stored, True otherwise
+            position_kp: the proportional factor of the position PID
+            position_ki: the integral factor of the position PID
+            position_kd: the derivative factor of the position PID
+            position_noise: the flat amount of noise added to the robot's phase on each localization step
+            heading_kp: the proportional factor of the heading PID
+            heading_ki: the integral factor of the heading PID
+            heading_kd: the derivative factor of the heading PID
+            heading_noise: ?
```
juliehelenek, line 56: we should probably just remove  this?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+            lf_ultrasonic: the ultrasonic at the front of the left side of the robot, used for boundary following
+            lb_ultrasonic: the ultrasonic at the back of the left side of the robot, used for boundary following
+            rf_ultrasonic: the ultrasonic at the front of the right side of the robot, used for boundary following
+            rb_ultrasonic: the ultrasonic at the back of the right side of the robot, used for boundary following
+            is_sim: False if the physical robot is being used, True otherwise
+            is_store: False if csv data should not be stored, True otherwise
+            position_kp: the proportional factor of the position PID
+            position_ki: the integral factor of the position PID
+            position_kd: the derivative factor of the position PID
+            position_noise: the flat amount of noise added to the robot's phase on each localization step
```
juliehelenek, line 52: also thinking about removing this too?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+            heading_kp: the proportional factor of the heading PID
+            heading_ki: the integral factor of the heading PID
+            heading_kd: the derivative factor of the heading PID
+            heading_noise: ?
+            phase: the phase of the robot
+            time_step: the amount of time that passes between each feedback loop cycle, should only be used if is_sim
+                is True
+            move_dist: the distance in meters that the robot moves per time dt
```
juliehelenek, line 60: is this also regardless of timestep?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.time_step = 1
+        self.position_kp = 1
+        self.position_ki = 0
+        self.position_kd = 0
+        self.position_noise = 0
+        self.heading_kp = 1
+        self.heading_ki = 0
+        self.heading_kd = 0
+        self.heading_noise = 0
+        self.move_dist = 0.5
```
juliehelenek, line 90: can we put these together with self.epsilon because they all refer to turning / turn radius?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.turn_angle = 3
+        self.turn_angle = self.turn_angle / self.time_step # dividing by time_step ignores the effect of time_step on absolute
```
juliehelenek, line 92: this doesn't make sense to me D: 
also, why not just do self.turn_angle /= timestep or self.turn_angle = 3 / self.time_step?

based on this, it seems like time_step is not just for simulations?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+            heading_ki: the integral factor of the heading PID
+            heading_kd: the derivative factor of the heading PID
+            heading_noise: ?
+            phase: the phase of the robot
+            time_step: the amount of time that passes between each feedback loop cycle, should only be used if is_sim
+                is True
+            move_dist: the distance in meters that the robot moves per time dt
+            turn_angle: the angle in radians that the robot turns per time dt regardless of time step
+            plastic_weight: the weight of the trash the robot has collected
+            init_threshold (Double): Radius from initial position that will detect robot is back in initial position
```
juliehelenek, line 63: float, not double (rip java)

juliehelenek, line 63: this comment isn't clear to me, just by reading it I'm not sure what it's supposed to be doing

juliehelenek: change this for all comments ^

engine/robot_state.py
juliehelenek, line 60: how is this a constant? doesn't this depend on the material we're driving on and the incline of the surface?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.heading_kd = 0
+        self.heading_noise = 0
+        self.move_dist = 0.5
+        self.turn_angle = 3
+        self.turn_angle = self.turn_angle / self.time_step # dividing by time_step ignores the effect of time_step on absolute
+        self.init_threshold = 1
+        self.goal_threshold = 1
+        self.noise_margin = 1
+        self.front_sensor_offset = 0  # TODO: replace this with how far offset the sensor is to the front of the robot
+        self.max_sensor_range = 600
```
juliehelenek, line 97: this is not in the specification above - what unit is this?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.sensor_measuring_angle = 75
```
juliehelenek, line 98: this is not in the specification above either

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.width_margin = 1  # TODO: replace this with actual margin
```
juliehelenek, line 99: not in specification - what does this refer to @_@

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.threshold_distance = ((self.width + self.width_margin) / 2) / math.cos(math.radians((180 - self.sensor_measuring_angle) / 2))
+        self.detect_obstacle_range = min(self.threshold_distance, self.max_sensor_range)  # set ultrasonic detection range
```
juliehelenek, line 101: im going to stop making comments about this, make sure that ALL attributes are in the specification + the order of them are consistent with between the specification + definitions 

also, try to group them 
i.e all sensor attributes together, all turning / moving attributes together : you're doing this for the most part but one or two are sticking out and it makes this a little more confusing

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # Measurements
+        self.state = np.array([[x_pos], [y_pos], [heading]])
```
juliehelenek, line 104: i dont think x_pos, y_pos, and heading need to be arrays, i think this should be (3,) (i.e np.array([x_pos, y_pos, heading])

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.truthpose = np.transpose(np.array([[x_pos], [y_pos], [heading]]))
+        self.plastic_level = 0
+        self.battery = 100
+        self.acceleration = [0, 0, 0]
+        self.magnetic_field = [0, 0, 0]
+        self.gyro_rotation = [0, 0, 0]
+        self.init_gps = (0, 0)
```
juliehelenek, line 111: what's the difference between init_gps and gps_data?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.gps_data = (0, 0)
+        self.imu_data = None  # will be filled by execute_setup
+        self.linear_v = 0
```
juliehelenek, line 114: is linear_v and angular_v getting set while we do the traversal? off the top of my head, I don't think we use these variables in the traversal phase

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.front_sensor_offset = 0  # TODO: replace this with how far offset the sensor is to the front of the robot
+        self.max_sensor_range = 600
+        self.sensor_measuring_angle = 75
+        self.width_margin = 1  # TODO: replace this with actual margin
+        self.threshold_distance = ((self.width + self.width_margin) / 2) / math.cos(math.radians((180 - self.sensor_measuring_angle) / 2))
+        self.detect_obstacle_range = min(self.threshold_distance, self.max_sensor_range)  # set ultrasonic detection range
+
+        # Measurements
+        self.state = np.array([[x_pos], [y_pos], [heading]])
+        self.truthpose = np.transpose(np.array([[x_pos], [y_pos], [heading]]))
```
juliehelenek, line 105: i don't understand why we have self.state and self.truthpose; if we needed to use self.truthpose, could we not just do self.state.T?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.init_gps = (0, 0)
+        self.gps_data = (0, 0)
+        self.imu_data = None  # will be filled by execute_setup
+        self.linear_v = 0
+        self.angular_v = 0
+        self.dist_to_goal = 0
+        self.prev_phase = self.phase
+        self.goal_location = (0, 0)
+       
+        # Sensors
```
juliehelenek, line 120: why is this the only section with a comment 😭do them all like this ^^

engine/robot.py
juliehelenek: i think they added it so we have the logic / know how to find the heading from the imu, but we might want to move it in the future

engine/robot.py
```diff
@@ -1,45 +1,14 @@
 import threading
```
juliehelenek, line 1: just an overall comment, but we shouldn't have files that are like 500+ lines, im wondering if there's a way we can split this up more? like maybe have a file for like robot_commands that handles moving forward etc. 

idk but this file seems tooo clunky

engine/robot_state.py
juliehelenek: looking at robot, it seems like this is an accumulator of some sort?

juliehelenek: oh, it's used to plot stuff - let's add that to the specification

engine/sim_trajectory.py
```diff
@@ -73,10 +74,9 @@ def get_plot_boundaries(nodes, delta):
             (CSV_PATH + '/rpi_to_gui_simulation.csv'), "a")
 
     # ser = serial.Serial("/dev/cu.usbserial-017543DC", 57600) # RPI_GUI_TEST
-    r2d2 = Robot(0, 0, math.pi / 4, epsilon=0.2, max_v=0.5,
-                 radius=0.2, init_phase=Phase.TRAVERSE, 
-                 is_sim=is_sim, is_store=is_store)
-
+    r2d2_state = Robot_State(x_pos=0, y_pos=0, heading=math.pi / 4, epsilon=0.2, max_velocity=0.5, radius=0.2)
```
juliehelenek, line 77: the initial heading should probably be read by the imu? why is this pi?

tests/compilation_tests/database_test.py
```diff
@@ -12,52 +12,59 @@
...
+                          radius=0)
+    robot_state_initial.plastic_level = 2
+    robot_state_initial.move_dist = 0.6
+    robot_state_initial.position_noise = 0.23
+    robot_state_initial.position_kp = 0.12
+    robot_state_initial.position_ki = 1.7
+    robot_state_initial.position_kd = 9.2
+    robot_state_initial.heading_kp = 0.2
+    robot_state_initial.heading_ki = 0.4
+    robot_state_initial.heading_kd = 0.16
```
juliehelenek, line 32: yeah... this is clunky - let's change it

tests/compilation_tests/database_test.py
```diff
@@ -12,52 +12,59 @@
...
+    robot_state_initial.is_sim = False
+    robot_initial = Robot(robot_state=robot_state_initial)
 
     db_initial = DataBase(robot_initial)
 
-    robot_one_param = Robot(x_pos=0, y_pos=0, heading=0, epsilon=0,
-                            max_v=0, radius=0, plastic_weight=3)
-    robot_one_param.battery = 46
-    robot_one_param.acceleration = [0.5, 0.2, 0.3]
+    robot_state_one_param = Robot_State(x_pos=0, y_pos=0, heading=0, epsilon=0, max_velocity=0,radius=0)
```
juliehelenek, line 44: what does one_param refer to?

tests/compilation_tests/robot_test/execute_setup_test.py
```diff
@@ -72,5 +70,16 @@ def test_radio_session_setup(self):
         robot_radio_session = RadioModule(None)
         self.assertEqual(robot_radio_session.connected, False)
 
+    #test motor_controller_setup calls the motor controller setup function, which will 
+    #print a series of commands. Make sure they are in the order of: 
+    #'go_forward', 'turn_left', 'turn_right', 'reverse', 'stop'
```
juliehelenek, line 75: is this a manual check? the assertEqual is not checking for this

tests/compilation_tests/robot_test/traverse_standard_test.py
```diff
@@ -16,7 +17,9 @@
 class TestTraverseStandardFunctions(unittest.TestCase):
 
     def test_init(self):
```
juliehelenek, line 19: i am confused on how these tests even pass - are they doing anything?

juliehelenek (changes requested):
made a few comments!~

engine/robot.py
```diff
@@ -306,29 +183,27 @@ def turn_to_target_heading(self, target_heading, allowed_heading_error, database
...
+                self.robot_state.state = self.update_ekf_step()
+            theta_error = target_heading - self.robot_state.state[2]
             w = self.head_pid.update(theta_error)  # angular velocity
-            _, limited_cmd_w = limit_cmds(0, w, self.max_velocity, self.radius)
+            _, limited_cmd_w = limit_cmds(0, w, self.robot_state.max_velocity, self.robot_state.radius)
 
-            if self.is_sim:
-                self.travel(0, self.time_step * limited_cmd_w)
-            else:
-                self.motor_controller.spin_motors(limited_cmd_w, 0)
```
AC408, line 325: if we're removing this, can we add it to travel?

engine/phase.py
```diff
@@ -0,0 +1,13 @@
...
+    """
+    An enumeration of different robot phases.
+    """
+    SETUP = 1
+    TRAVERSE = 2
+    AVOID_OBSTACLE = 3
+    RETURN = 4
+    DOCKING = 5
+    COMPLETE = 6
+    FAULT = 7
```
AC408, line 13: might not be relevant to this pr, but I remember we might be having a phase for manual control?

engine/robot_state.py
AC408: mb, this is used for obstacle avoidance. we say the robot completes a cycle (and therefore goal untraversable) if the robot left the init_threshold and came back to it

engine/robot_state.py
AC408: true, will change in separate pr

engine/robot_state.py
AC408: also apparently,  turn_angle = the angle in radians that the robot turns per time dt regardless of time step. I think it has something to do with Roomba mode, will investigate later

AC408: by "it", I mean dividing by tilmestep

engine/robot_state.py
AC408: front_sensor_offset and max_sensor_range in cm, can prob be moved into the obstacle avoidance thread instead of being a class attribute bc only rly being used in that function/thread. idk, which makes more sense?

engine/robot_state.py
AC408: margin on the side of the robot so when doing boundary following, robot doesn't follow the wall too closely. can standardize to cm

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        self.turn_angle = 3
+        self.turn_angle = self.turn_angle / self.time_step # dividing by time_step ignores the effect of time_step on absolute
+        self.init_threshold = 1
+        self.goal_threshold = 1
+        self.noise_margin = 1
+        self.front_sensor_offset = 0  # TODO: replace this with how far offset the sensor is to the front of the robot
+        self.max_sensor_range = 600
+        self.sensor_measuring_angle = 75
+        self.width_margin = 1  # TODO: replace this with actual margin
+        self.threshold_distance = ((self.width + self.width_margin) / 2) / math.cos(math.radians((180 - self.sensor_measuring_angle) / 2))
```
AC408, line 100: cm, how far the away the obstacle can be such that we will detect an obstacle (assuming infinite sensor range). by detecting an obstacle, I mean obstacle x position within width of the robot + margin (assuming width = x).

engine/robot_state.py
AC408: cm, how far the away the obstacle can be such that we will detect an obstacle (taking into account threshold_distance and physical sensor limitation)

engine/robot_state.py
AC408: In turn, angle changes by (turn_angle * self.time_step), so when we want the robot to turn 90 degrees but time_step is not 1, it won't turn 90 degrees. (which also doesn't align with the comment in turn) I think dividing by time_step here makes it so that the actual radians turned is 3 (not sure why its that high lol). I think I kept the two lines separate to show that the actual angle turned when using turn_angle is 3 instead of turn_angle/time_step, which could be confusing

AC408 (changes requested):
made some comments. made some replies as well. mostly think the changes here are fine and that most of the comments about changes should be done in another pr, since this pr is mostly to change all attributes of robot class to robot_state

tests/compilation_tests/database_test.py
AdityaKompella6: Maybe we can use **kwargs for the class init so we don't have to assign everything using the dot syntax

AdityaKompella6 (commented):
Good Job overall lots of files changed wooooo

Shreya1233 (commented):
Gen comment: why is movement_test and gps_test not included in the test_runner since the way we make a BasicMotorController object is changed now?

engine/robot.py
sazhang02: ok! lets leave this for now then

electrical/gps.py
```diff
@@ -4,13 +4,14 @@
...
+from engine.is_raspberrypi import is_raspberrypi
+if is_raspberrypi():
     from ublox_gps import UbloxGps
 
 
 class GPS:
-    def __init__(self, init_serial, is_sim):
-        self.is_sim = is_sim
+    def __init__(self, init_serial):
+        self.is_sim = not is_raspberrypi()
```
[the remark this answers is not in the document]
sazhang02: Oops yes! aayush brought this up and I forgot to change it to be passed down from robot_state (already has is_sim flag). Good catch o7

Also going to note here that we still need is_raspberrypi for the import statement since the flag set in the init statement

tests/functionality_tests/obstacle_tracking_test.py
```diff
@@ -182,12 +186,14 @@ def test_min_value(self):
...
-                             front_ultrasonic=None, lb_ultrasonic=None, lf_ultrasonic=None, rb_ultrasonic=None,
-                             rf_ultrasonic=None)
+        wide_robot_state = Robot_State(x_pos=0, y_pos=0, heading=math.pi/2, epsilon=0.2, max_velocity=0.5, radius=0.2)
+        wide_robot_state.phase = 2
+        wide_robot_state.width = 800
+        wide_robot = Robot(wide_robot_state)
+        skinny_robot_state = Robot_State(x_pos=0, y_pos=0, heading=math.pi/2, epsilon=0.2, max_velocity=0.5, radius=0.2)
+        skinny_robot_state.phase = 2
+        skinny_robot_state.width = 200
+        skinny_robot = Robot(skinny_robot_state)
```
AC408, line 196: I think the only thing that is needed is changing the width? So maybe smthg like wide_width = 800, skinny_width = 200, robot_state.width = skinny_width or smthg like that to unduplicated

engine/database.py
```diff
@@ -16,21 +16,21 @@ def __init__(self, robot):
             heading_pid: heading PID in format of [proportional factor, integral factor, derivative factor]           
             move_dist: the distance in meters that the robot moves per time dt
             turn_angle: the angle in radians that the robot turns per time dt regardless of time step 
-            plastic_weight: the plastic_weight of the trash the robot has collected
+            plastic_level: the plastic_level of the trash the robot has collected
```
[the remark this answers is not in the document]
sazhang02: I think %'s based on the break beam sensors. Since we have 16 w 4 for each level ill add 25, 50, 75, 100 for now and a TODO to check in the future when electrical has a better idea of how its going to work.

engine/database.py
sazhang02: does this mean we don't want to have a heading pid then?

engine/robot_state.py
sazhang02: Just for context, all the attributes in `robot_state` are copied over from what currently exists in our initialization of `robot`.

I think radius and width was added by Allen for obstacle avoidance but I'm not too sure. @AC408 if u have any input?

engine/robot_state.py
sazhang02: Same as above (robot_state + spec is what we had in robot's initialization).

Since epsilon seemed to have been defined a long time ago (maybe before I joined??), I'm not too sure how it should be defined. If we want to reevaluate our current attributes we can talk about it at subteam and make a separate PR for it (since this pr is more focused on the refactoring/moving the attributes from one place to another).

engine/robot_state.py
sazhang02: does this mean why do we need `avoid_obstacle` or other all of the 4 flags that are selected?

engine/robot_state.py
sazhang02: good point

engine/robot_state.py
AC408: I added width for obstacle avoidance, not sure what radius is. Might be able to remove width and use radius

engine/robot_state.py
sazhang02: let's further refactor this separately w the other attributes were re-evaluating!

engine/robot_state.py
sazhang02: same as above

engine/robot_state.py
sazhang02: same as above

engine/robot_state.py
sazhang02: same as above -- but also does this mean there's no heading pid?

engine/robot_state.py
sazhang02: same as above

engine/robot_state.py
sazhang02: Oh true. I put a section for user defined parameters but i think this makes more sense!

engine/robot_state.py
sazhang02: all 4 specs?

engine/robot_state.py
sazhang02: I reordered slightly based on what i understood from robot's attributes, but let's talk about this all together since it's a conglomeration of things added by old and new members over the years and idk everything about it

engine/robot_state.py
sazhang02: ok radius must've been made back in the day. so quick w the response Allen LOL

engine/robot_state.py
sazhang02: Do you mean `# Sensors`?? A bit confused bc I put comments at the top of each section (we should re-evaluate how we want to organize the attributes tho anyways)

engine/robot.py
sazhang02: agreed! ill add it to the list of new tasks -- think we could further refactor robot aside from splitting out these attributes to robot_state

engine/sim_trajectory.py
sazhang02: Hm not sure 😅 , but just reiterating that the purpose of this PR was to move the attributes out from Robot to Robot_State, so changes to other files are for the purpose of updating how initialization of Robot to match this refactoring (passing in Robot_State and moving initialization there instead).

It's a good point to question why we have been initializing things this way, just a separate issue.

tests/compilation_tests/database_test.py
sazhang02: Good idea!

sazhang02: @juliehelenek @AdityaKompella6 what do ug think about **args also for initializing the attributes for our robot in `Robot_State`? Im not too familiar w how **kwargs and **args work so wanna see what ug think

tests/compilation_tests/database_test.py
sazhang02: Same as before, just updating the initialization of Robot to reflect moving attributes to be held in Robot_State. But since I wrote this before w/ Angela I believe we intended this to be a robot representing the test case of a robot w only one parameter/attribute changed from its default initialization state (simplest case of whether the packet produced is properly updated attributes of robot change from the default initialization).

tests/compilation_tests/robot_test/execute_setup_test.py
sazhang02: Not sure (as I didn't write this just copy-pasted) -- to preface where this change came from, basically I noticed there were two copies of execute_setup_test when I went through all the test files to see where we were initializing Robot. I combined them since it didn't make sense to have two duplicate files, and added any lines that differed between them!

Again, separate task from the purpose of this PR, but a problem nonetheless. Added to list of ongoing new tasks that you have pointed out previously in the PR summary

tests/compilation_tests/robot_test/traverse_standard_test.py
sazhang02: same as execute_setup but I noticed there was 2 traverse_standard_test files and combined the duplicates + updating initialization of Robot to reflect the refactoring initialization (did not write this file).

engine/robot.py
sazhang02: oops! accidental deletion

sazhang02: added back o7

engine/phase.py
sazhang02: yus not really relevant bc phase was just moved into a separate file for reducing robot clutter but good point I will add it to our list of other things we should figure out

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+            heading_noise: ?
+            phase: the phase of the robot
+            time_step: the amount of time that passes between each feedback loop cycle, should only be used if is_sim
+                is True
+            move_dist: the distance in meters that the robot moves per time dt
+            turn_angle: the angle in radians that the robot turns per time dt regardless of time step
+            plastic_weight: the weight of the trash the robot has collected
+            init_threshold (Double): Radius from initial position that will detect robot is back in initial position
+            goal_threshold (Double): Threshold from goal that will be detected as reaching goal in obstacle avoidance
+            noise_margin (Double): Margin from init_threshold that the robot has to leave before detecting robot has
```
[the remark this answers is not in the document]
sazhang02: another pr!

engine/robot_state.py
sazhang02: is this something to add to the spec?

engine/robot_state.py
```diff
@@ -0,0 +1,138 @@
...
+        Instance Attributes:
+            epsilon: dictates the turning radius of the robot. Lower epsilon results in tighter turning radius.
+            max_velocity: the maximum velocity of the robot
+            radius: the radius of the robot
+            width: width of the robot in cm
+            front_ultrasonic: the ultrasonic at the front of the robot, used for detecting obstacles
+            lf_ultrasonic: the ultrasonic at the front of the left side of the robot, used for boundary following
+            lb_ultrasonic: the ultrasonic at the back of the left side of the robot, used for boundary following
+            rf_ultrasonic: the ultrasonic at the front of the right side of the robot, used for boundary following
+            rb_ultrasonic: the ultrasonic at the back of the right side of the robot, used for boundary following
```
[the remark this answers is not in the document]
sazhang02: hmm i think this relates to a todo i added just now about how were initializing our other sensors like GPS,IMU,... etc bc we do it both in our initialization and in execute_setup. Nonetheless, lets discuss and do this separately!

tests/functionality_tests/obstacle_tracking_test.py
sazhang02: a fast review

sazhang02: Not sure if this is correct with Python but if we share the robot_state, I ~think~ that they would share the same reference and any changes to the attribute width would affect both the wide_robot and skinny_robot's attributes?

For example, if wide_robot was initialized with robot_state and set to wide_width and then skinny_robot was initialized using that same robot_state, setting the attribute to skinny_width would change robot_state (and thus wide_robot.robot_state is skinny_width)

sazhang02:
@juliehelenek Love the enthusiasm about our code but just want to reiterate that A LOT of issues brought up are just current issues with our code that don't relate to what this PR is doing. I summed up the main other issues you brought up in the PR summary, and we should definitely look into them more (as they were very valid points). However, these are separate issues/tasks as this PR is just refactoring **where we put the robot's attributes** and updating how we initialize Robot (not how it works, why are our test cases like this, etc.). Think of this as separating out code from the long `robot.py` file. Even though robot_state is all green/added lines, please keep in mind that I just cut + pasted the `pre-existing` code in robot (I did not write/implement all of that, just mainly moved the attributes we are initializing manually to the top of the init) so any issues pointed out are with that original implementation should be addressed **separately**.

sazhang02:
While going through review comments I also realized I missed updating some initialization of `Robot` in other test files (especially the functionality tests). Fixed + pushed changes for this.

sazhang02:
@Shreya1233 good point bc they have sim flags 🤔 i think they should be in the functionality_tests folder actually (bc compilation_tests is for unit tests), which is why in the test_runner, which runs unit tests, they wouldn't do much (since it prints/manual check) atm. Added to list in PR summary of separate tasks that arose from everyone's comments.

engine/database.py
AC408: wait what. I think we need a heading pid. heading pid should be different from traversal pid

engine/robot_state.py
AC408: I think Julie is talking about the heading_noises?

engine/sim_trajectory.py
AC408: maybe bc sim so can't get it by reading imu bc no imu in sim?

engine/robot_state.py
AC408: somehow yeah. I don't think my description is concise enough to be used as spec

tests/functionality_tests/obstacle_tracking_test.py
AC408: im not sure about this as well

engine/database.py
sazhang02: i asked julie on slack and i think she meant testing for noise in general. Since database is going to be scrapped since this refactoring allows us to replace/rewrite it we can prob account for it later

sazhang02:
Updated to use **kwargs for initialization! also realized obs avoidance code in robot references xpos and ypos directly (but we want to keep this within state, so all updates are to xpos/ypos are centralized -- don't have to update xpos and state[0]).

engine/database.py
AC408: so are we still talking about heading pid or heading noise?

tests/compilation_tests/database_test.py
```diff
@@ -12,39 +12,25 @@
...
-    robot_state_initial.heading_kp = 0.2
-    robot_state_initial.heading_ki = 0.4
-    robot_state_initial.heading_kd = 0.16
-    robot_state_initial.phase = Phase.TRAVERSE
-    robot_state_initial.state = np.array([[10], [20], [50]])
-    robot_state_initial.battery = 98
-    robot_state_initial.magnetic_field = [0.1, 0.2, 0.3]
-    robot_state_initial.gyro_rotation = [0.5, 0.2, 0.6]
-    robot_state_initial.acceleration = [4.25, 3.2, 0.1]
-    robot_state_initial.is_sim = False
```
AC408: why did we change is_sim to false? is true a bug

AC408 (changes requested):
looks good, what's kwarg and what does it do?

AC408:
> looks good, what's kwarg and what does it do?

might it be possible to include that in the PR description?

juliehelenek (approved):
looks good (:

jinlin270 (approved):
This is actually really cool and decouples a lot of stuff. Everything looks good. Just wondering if it is ok that we have the is_sim flag at multiple places instead of it being passed down.

tests/compilation_tests/database_test.py
sazhang02: Yes o7 i realized that it if we set is_sim to False during the initialization it causes us to run the sensor initialization at the end of `Robot_State`. Since all of the variables in a packet hold relatively the same meaning I should be ok to change the is_sims to True!

sazhang02:
> looks good, what's kwarg and what does it do?

I think it's a Python way to initialize multiple variables. Vaguely remembered from intro to backend when Aditya pointed it out.
https://github.com/cornellnexus/src/pull/112#discussion_r1131614806
https://www.geeksforgeeks.org/args-kwargs-python/

tests/functionality_tests/obstacle_tracking_test.py
sazhang02: hmm i think its objects are passed by reference since they're mutable?
https://www.tutorialspoint.com/how-are-arguments-passed-by-value-or-by-reference-in-python

sazhang02:
> This is actually really cool and decouples a lot of stuff. Everything looks good. Just wondering if it is ok that we have the is_sim flag at multiple places instead of it being passed down. It looks like we have separate is_sim flags in robot_state, motor_controller, and GUI.

@jinlin270 Current plan is is_sim initialized in Robot_state and reinitialized to only electrical files (like motor_controller) since we cannot pass the `robot_state` directly to them (since the electrical objects like motor_controller are initialized in robot_state + creates circular import dependency) -- otherwise uses robot_state for is_sim. Besides electrical files -- sim_traj and GUI should be ok to have them since sim_traj will be the file that creates robot state (and thus doesn't have the is_sim flag readily available to reference) and GUI have robot_state available since its on our laptops (+ existing files going to be scrapped anyways LOL). I think those are the only ones but if you notice it somewhere else lmk and ill change!

jinlin270:
> looks good, what's kwarg and what does it do?

tests/functionality_tests/obstacle_tracking_test.py
sazhang02: ok update i ran it with using one state and it passed BUT i realized w kwargs it'll prob make more sense readability-wise to have individual states for them

AC408 (approved):
looks good to me

Shreya1233 (approved):
It looks good to merge into main!

PR #117, opened 2023-03-09
Adding break beam sensor code

AmberL235:
## References
Code was based on snipets found here: https://simonprickett.dev/using-a-break-beam-sensor-with-python-and-raspberry-pi/

### Description
Original code printed out to system when it detected a beam break. Modified code so that it returns a boolean beam_broken. Print statements left in for testing and inputting an "enter" keypress will stop the detection. 

## Test Plan
Ran code on RPi with the sensors hooked up and tested by moving my hand between the sensors to block the beam. Compared printed outputs to my blocking patterns and confirmed they were correct.


electrical/RPi BreakBeam.py
```diff
@@ -0,0 +1,24 @@
...
+beam_broken = False
+
+def break_beam_callback(channel):
+    if GPIO.input(BEAM_PIN):
+        beam_broken = False
+        print(beam_broken)
+    else:
+        beam_broken = True
+        print(beam_broken)
+    return beam_broken
```
Shreya1233, line 13: This can be condensed into a few lines! 
if not GPIO.input(BEAM_PIN):
         beam_broken = True
print(beam_broken)
return beam_broken

electrical/RPi BreakBeam.py
```diff
@@ -0,0 +1,24 @@
+import RPi.GPIO as GPIO
+
```
sazhang02, line 2: hi

electrical/RPi BreakBeam.py
nicholaspapapanou: I agree!

electrical/RPi BreakBeam.py
```diff
@@ -0,0 +1,24 @@
+import RPi.GPIO as GPIO
+
+BEAM_PIN = 17
+beam_broken = False
+
+def break_beam_callback(channel):
+    if GPIO.input(BEAM_PIN):
+        beam_broken = False
+        print(beam_broken)
```
sazhang02, line 9:
```suggestion
        # print(beam_broken)

```
Maybe we can comment these out and uncomment them when testing?

electrical/RPi BreakBeam.py
```diff
@@ -0,0 +1,24 @@
...
+BEAM_PIN = 17
+beam_broken = False
+
+def break_beam_callback(channel):
+    if GPIO.input(BEAM_PIN):
+        beam_broken = False
+        print(beam_broken)
+    else:
+        beam_broken = True
+        print(beam_broken)
```
sazhang02, line 12:
```suggestion
        # print(beam_broken)

```
see above

Andrewe0616 (approved):
Looks good to me!

sazhang02 (changes requested):
Nice PR!!! 🥳  Not sure about the above change.

electrical/RPi BreakBeam.py
sazhang02, line 1: can we remove the space in the file name?

sazhang02 (changes requested):
example

electrical/RPi BreakBeam.py
nicholaspapapanou: I second this!

nicholaspapapanou (changes requested):
Great job Amber! Only change is Sammie's suggestion to remove the space in the file name.

sazhang02 (approved):
Looks good to me!!! Thank you for resolving comments + letting us review tg td xD

Good job re-requesting reviews too. Well done 🥳

nicholaspapapanou (approved):
Awesome job, Amber!

electrical/break_beam.py
```diff
@@ -0,0 +1,26 @@
...
+    return beam_broken
+
+def check_break_beam():
+    GPIO.setmode(GPIO.BCM)
+    GPIO.setup(BEAM_PIN, GPIO.IN, pull_up_down=GPIO.PUD_UP)
+    GPIO.add_event_detect(BEAM_PIN, GPIO.BOTH, callback=break_beam_callback)
+
+    message = input("Press enter to quit\n\n")
+    GPIO.cleanup()
+    return break_beam_callback
```
juliehelenek, line 24: I don't think you're calling the function here - when you call a function in Python, you need the parenthesis / parameter, i.e return break_beam_callback(BEAM_PIN) or smthg

juliehelenek (changes requested):

electrical/break_beam.py
nicholaspapapanou: I'm actually not sure if returning break_beam_callback is even necessary here. It seems that after you add_event_detect, break_beam_callback will automatically be run whenever BEAM_PIN's state changes.

nicholaspapapanou (changes requested):
We can test tomorrow to see whether removing the return break_beam_callback from check_break_beam still gives the desired functionality. The reason for the weird calling syntax might be because we aren't actually meant to call it ourselves!

electrical/RPi BreakBeam.py
juliehelenek: another cool simplification: 

`beam_broken = not GPIO.input(BEAM_PIN)
print(beam_broken)
return beam_broken`

electrical/break_beam.py
```diff
@@ -0,0 +1,26 @@
...
+def break_beam_callback(channel):
+    if GPIO.input(BEAM_PIN):
+        beam_broken = False
+        # print(beam_broken)
+
+    else:
+        beam_broken = True
+        # print(beam_broken)
+
+    return beam_broken
```
Andrew-Lowitt, line 15: You can compress this if statement into a single line as seen in the example below.
```suggestion
   
    return not GPIO.input(BEAM_PIN)

```

juliehelenek (approved):

PR #118, opened 2023-03-16
Added Motor Estimation Equations to EKF 

juliehelenek:
## References
[Documentation of EKF and where equations come from](https://docs.google.com/document/d/1U4ov8-be8NQDjB8Fb7kxxulbnkHCoAF7c8RZAqNis8k/edit#)
### TLDR: Replace this with 1-3 sentence TLDR
Previously, we didn't have a completed predict step because we didn't have a function to get our controls. Since the EKF is broken up into two steps, predict and update, this was a pretty major part that we didn't have
### Description
The derivation of the equations is detailed further in the documentation, but pasted here again for convenience. 
image
image
image
image
image
image

## Test Plan
This was tested by running localization_visualizer with the following line:
(live 0: the data is NOT live)
(data 2: using BOTH GPS and IMU data)
(ekf 1: using the EKF)
python -m tests.functionality_tests.localization_visualizer live=0 data=2 ekf=1

This file **does not use a robot object** but is dealing with the state directly. In other words, we're only focusing on "Does the EKF change x, y, and pos?" This design choice was made a few years ago, so it honestly could be updated but I kept it as is for now.


engine/ekf.py
```diff
@@ -101,14 +103,61 @@ def get_h_jac(self, pose):
         """
         return np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
 
-    def predict_step(self, control):
+    def get_arc_lengths(self, left_side_w, right_side_w):
```
AC408, line 106: is this under the assumption that the time step is 1?

engine/ekf.py
```diff
@@ -101,14 +103,61 @@ def get_h_jac(self, pose):
...
+        Given from electrical: arc_lengths: a tuple such that the first entry is the 
+        left-side arc distance and the second entry is the right-side arc distance
+        
+        Arc distance is calculated from motor_velocity * timestep
+        """
+
+        L_L = arc_lengths[0]
+        L_R = arc_lengths[1]
+
+        try:
```
AC408, line 137: What if L_R = L_L?

tests/functionality_tests/localization_visualizer.py
```diff
@@ -180,8 +185,9 @@ def animate(i):
                 new_location = (measurements[0], measurements[1])
 
             elif not is_live and use_ekf:
-                mu_bar = ekf.mu  # temporary placeholder
-                sigma_bar = ekf.sigma  # temporary placeholder
+                arc_lengths = ekf.get_arc_lengths(1,1) # temporary placeholder
```
AC408, line 188: is this something we can vary and test?

tests/functionality_tests/localization_visualizer.py
```diff
@@ -31,9 +31,9 @@ class DataType(IntEnum):
...
     zone_photo = "geo_images/engineering_quad.png"
-    imu_data_file = "csv/imu_360_sample1.csv"
-    gps_data_file = "csv/GPS_13-11-2021_14-18-15.txt"
-    # gps_data_file = "csv/GPS_22-11-2021.txt"
+    imu_data_file = "./csv_files/imu_360_sample1.csv"
+    # gps_data_file = "./csv_files/GPS_13-11-2021_14-18-15.txt"
+    gps_data_file = "./csv_files/GPS_22-11-2021.txt"
 
     # Read command-line arguments
     if len(sys.argv) != 4 or "live=" not in sys.argv[1] or "data=" not in sys.argv[2] or "ekf=" not in sys.argv[3]\
```
AC408, line 39: how do you pass in the args? I'm confused. Maybe an example in comments?

engine/ekf.py
juliehelenek: im not really sure how to incorporate the timestep - ive seen it used in other areas of the code - im not sure if its the same throughout or if each one has their own time-cycle --- open to suggestions here

engine/ekf.py
juliehelenek: if L_R = L_L, i set r = 0 to avoid division by 0 - since r is the distance from the robot to center of curvature, i think 0 would make sense

tests/functionality_tests/localization_visualizer.py
juliehelenek: which args? the data files? they're CSVs and we're just reading one line per iteration of forloop 
we're only working with gps/imu data - not with the robot at all (this file goes back to when we didnt even have a robot to work with)

tests/functionality_tests/localization_visualizer.py
juliehelenek: these arc lengths are dependent on the velocity - im thinking that this is another piece of data that we would have to collect or simulate (im not sure how we would go about simulating it though)

AC408: I think as long as we have a theta, r, and width, we have a valid arclength (arclength = theta*R, where R = r or r + width. Then we can randomize theta, r, and width to see if the property holds

tests/functionality_tests/localization_visualizer.py
AC408: like, how would you run this test file?

engine/ekf.py
AC408: I think timestep is the cycle time, so effectiively how long we wait before changing the input voltage to the motors (in the while loop there's a delay of 20ms and changing input voltage to motor changes once per while loop iteration? so I think that'll mean the time step is 20ms). Even if tilmestep changes, it might be a good option to have it as a param and make the user define it

engine/ekf.py
AC408: sg

tests/functionality_tests/localization_visualizer.py
```diff
@@ -113,6 +113,11 @@ class DataType(IntEnum):
             frames = min(len(imu_readings) - 1, len(gps_readings) - 1)
 
     # EKF setup
+
+    # First GPS reading -> 0,0 in global frame 
+    # Map GPS coordinates to global 
+    # MAp GPS initial to 0,0
+    # New measurement - initial -> convert to meters
```
sazhang02, line 120:
```suggestion
    # First GPS reading -> 0,0 in global frame 
    # Map GPS coordinates to global 
    # Map GPS initial to 0,0
    # New measurement - initial -> convert to meters 
```
Also is this an overview of the below code under `if use_ekf`? If so, maybe add another comment on top noting the context of these steps like "#Calculate initial mu and sigma for EKF initialization" (not sure if this is right)?

sazhang02 (approved):
Nice pr! Added some minor comments and piggybacking on allen's review. In Julie's math we trust o7 Lgtm

engine/ekf.py
juliehelenek: yeah the name was shortened on the documentation cuz it would be annoying to add into all the equations but i can update this

tests/functionality_tests/localization_visualizer.py
juliehelenek: making a note of this for potential testing

engine/ekf.py
```diff
@@ -101,14 +103,61 @@ def get_h_jac(self, pose):
...
+    def get_arc_lengths(self, left_side_w, right_side_w):
+        """
+        Returns the arc_lengths of the left side of the robot and the right side of the robot 
+        given the angular velocity of the left side and angular velocity of the right side. 
+
+        left_side_w: angular velocity of the motors on the left side of the robot
+        right_side_w: angular velocity of the motors on the right side of the robot
+        """
+        # TODO: from electrical: from motors, get angular velocity
+        # TODO: from angular velocity, find arc_lengths
```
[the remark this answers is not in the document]
juliehelenek: these are changed!

engine/ekf.py
```diff
@@ -101,14 +103,61 @@ def get_h_jac(self, pose):
...
+
+
+    def get_controls(self, arc_lengths): 
+        """
+        Returns phi, the angle changed in a timestep and d, the displacement traveled in a timestep.
+
+        Given from electrical: arc_lengths: a tuple such that the first entry is the 
+        left-side arc distance and the second entry is the right-side arc distance
+        
+        Arc distance is calculated from motor_velocity * timestep
```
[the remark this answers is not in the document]
juliehelenek: done! arc distance isnt going to be calculated in this way anymore lol

engine/ekf.py
```diff
@@ -101,14 +103,61 @@ def get_h_jac(self, pose):
...
+        """
+
+        L_L = arc_lengths[0]
+        L_R = arc_lengths[1]
+
+        try:
+            r = L_L * self.robot_width / (L_R - L_L)
+        except:
+            r = 0
```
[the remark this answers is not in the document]
juliehelenek: makes sense!

engine/ekf.py
```diff
@@ -123,17 +172,13 @@ def update_step(self, mu_bar, sigma_bar, measurement):
...
-             inv(jac_H * sigma_bar * np.transpose(jac_H) + self.R))
-        expected_measurement = self.get_expected_measurement(mu_bar)
-
-        # print("kalman_gain")
-        # print(kalman_gain)
-        # print("measurement")
-        # print(measurement)
-        # print("expected_measurement")
-        # print(expected_measurement)
+        # K = H * sigma_bar * H.T * inv(H * sigma_bar * H.T + R)
```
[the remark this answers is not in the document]
juliehelenek: deleted

tests/functionality_tests/localization_visualizer.py
juliehelenek: added!

engine/ekf.py
```diff
@@ -103,20 +103,20 @@ def get_h_jac(self, pose):
...
         """
-        # TODO: from electrical: from motors, get angular velocity
-        # TODO: from angular velocity, find arc_lengths 
+        # TODO: from electrical: from motors, number of revolutions
+        # TODO: from number of revolutions, using the circumference of the wheels, figure out the distance traveled 
         # Note: this was put on hold Spring '23 because Professor advised that getting velocity
         # in this fashion was not that accurate 
-        left_arc_length = left_side_w # TEMP
-        right_arc_length = right_side_w # TEMP
+        left_arc_length = left_motor_revolutions # this will be some sort of relation with the wheel circumference
```
AC408, line 118: I believe its (revolution since last call) * 2 * pi * wheel radius

engine/ekf.py
juliehelenek: resolving this conversation - timestep will be determined by the caller

AC408 (approved):

Shreya1233 (approved):
I don't really have any feedback! Everything looks good!

engine/ekf.py
juliehelenek: will keep this in mind when communicating w electrical for this function!

PR #120, opened 2023-03-18
April tag testing

AdityaKompella6:
Docking Script Steps 1-4: 
## Summary
These are the first four steps of our docking algorithm. The purpose of these four steps is to have our central camera align with the central April tag of a side of the docking station. The robot is a certain, fixed distance from the April tag. When running the script you will see a popup that shows the angle, distance, and depth of the tags. Only look at the file docking_script.py. The rest are images of generated April tags, camera calibration files, or files with specific functionalities that we fully integrated in our docking script. 
### Description
https://docs.google.com/document/d/1ejBQGuzowJVMsnG142-7l5erR4W63vwy2U8ZFZIUtxI/edit
Changes: 
We have a function distance_to_camera that calculates the perpendicular distance of the center of the camera to the center of an April tag. 
We have a function direction_of_tag that determines if the camera is viewing the tag from the right or left side of the tag.  
We have a dictionary idDict that maps the tag ids of April Tags of the 36h11 family to string names. For example, an April tag id 1 is mapped to R1. 
We defined the motor to be a BasicMotorController instance and set is_sim to true for simulated testing purposes. 
We have a while loop that continuously processes frames from the camera and executes the steps in order.
To terminate the loop Step 4 has to finish or you press "q".
If we don't detect any corners and if we don't see anything for ten seconds we want the robot to spin for 15 seconds in a circle.
If we do see April tags for the first tag we see we calculate the angle, horizontal distance(Middle of the tag to camera center), and perpendicular distance.
If the tag we see is not the center tag, we first determine what direction of the center tag we are and we turn in the opposite direction until we align ourselves with the center tag. We then go forward until we can't see any tags. 
Now we finish Step 1 and go to Step 2.
We reverse until we can see a tag.
We are done with step 2.
Now in Step 3, we turn until we are straight on with the tag. We are making the assumption that the robot turns 3 degrees per second. We divide the current angle by 10 to calculate how many times the robot should turn until it's straight onto the tag. 
We reverse by a fixed amount in Step 4.
We are done:)
## Test Plan
We did simulated testing with existing cameras and April tags on a white or light-colored surface. This works because we simulated it. We found out constants such as the depth_min and angle_min (minimum distance between center of camera and center of April tag) and angle (minimum angle between tag's center and camera's center that the camera) the camera that can be detected by the camera and other important details by testing it in the real world. We manually made sure that the code was entering the right step at the right time by looking at the console output. To run the code run python -m april_tag_updates.docking_script

##TODO:
~~1. Finish Steps 1-4 with Basic Motor Controller Class + Simulated Testing - Done~~
2. Physically Test the Robot to accurately determine numbers especially for Steps 3-5 
3. Implement step 5
4. 3. Replace movement methods with those in the Motor Controller Class.
5. Integrate with robot.py under execute_docking and replace movement methods with methods in the Robot class.


april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
+expected_dist = 37
```
AC408, line 1: expected distance to/of what? in what units?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+window = 'Camera'
+cv2.namedWindow(window)
+def distance_to_camera(knownWidth, focalLength, perWidth):
+  # compute and return the distance from the maker to the camera
+  return (knownWidth * focalLength) / perWidth
+
+def direction_of_tag(corner,image):
+  (_, _, _, _) = corner
+  _,w = image.shape[0], image.shape[1]
+  cx, _ = int(960),int(540)
```
AC408, line 22: is this based on how large the window is? would it be more robust to find the width and length of the window and dividing that by 2?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+  if  bottomLeft[0] > cx:
```
AC408, line 23: perhaps we could pass this as a parameter? It might not be clear otherwise where this is from

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+    return "right"
+  else:
+    return "left"
+
+idDict = {1: "R1", 2: "R2", 3: "R3"}#TODO add all the other tags
+visited = []
+timeout =  10
+angle_zero = 10
```
AC408, line 31: units? what's angle_zero?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+
+while True:
+    if start_time is None:
+        start_time = time.time()
+    ret, image = cap.read()
+    (corners, ids, rejected) = cv2.aruco.detectMarkers(image, arucoDict, parameters=arucoParams)
+    if len(corners) == 0:
+      if time.time() - start_time > timeout and step1_done is False:
+        #Step 1 - Check for april tags rn by rotating in circle for 15 seconds.
+          motor.turn_right()
```
AC408, line 47: how does this guarantee that the robot rotates in a circle for 15 seconds? Maybe do some time.sleep(1500) or smthg?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+motor = BasicMotorController(True)
+start_time = None
+
+while True:
+    if start_time is None:
+        start_time = time.time()
+    ret, image = cap.read()
+    (corners, ids, rejected) = cv2.aruco.detectMarkers(image, arucoDict, parameters=arucoParams)
+    if len(corners) == 0:
+      if time.time() - start_time > timeout and step1_done is False:
```
AC408, line 45: is timeout arbitrary? or has the value for timeout been found experimentally?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+        #Step 1 - Check for april tags rn by rotating in circle for 15 seconds.
+          motor.turn_right()
+      if step1_done and step2_done is False:
+        #Step 2 - go back until len(corners) = 1 or when we see at least 1 april tag
+        motor.reverse()
+    if len(corners) > 0:
+      if step1_done:
+        step2_done = True
```
AC408, line 53: this logic doesn't make sense to me. If step1 is done, step2 is done? What's the point of having step2 then?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+      start_time = time.time()
+      for (corner, markerId) in zip(corners, ids):
+          markerId = float(markerId)
+          (topLeft, topRight, bottomRight, bottomLeft) = corner[0]
+          pixel_width = np.sqrt((bottomRight[0] - bottomLeft[0])**2 + (bottomRight[1] - bottomLeft[1])**2)
+          depth = distance_to_camera(known_width,focal_length,pixel_width)
+          cxt,cyt = int((topLeft[0] + bottomRight[0])//2),int((topLeft[1] + bottomRight[1])//2)
+          cx, cy = int(960),int(540)
```
AC408, line 61: maybe we can make this global?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+          cv2.circle(image, (cxt, cyt), radius = 20, color = (0, 0, 255), thickness = -1)
+          cv2.circle(image, (cx, cy), radius = 20, color = (0, 0, 255), thickness = -1)
+          cv2.line(image,(cxt, cyt),(cx, cy), color = (0, 0, 0), thickness = 2)
+          ratio= known_width/pixel_width
+          dpix = np.sqrt((cx-cxt)**2+(cy-cyt)**2)
+          horiz_dist = dpix * ratio
```
AC408, line 67: Doesn't this find the actual (physical, real world) distance between the center and the center of the April tag?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+          midpoint = (int((cx + cxt) //2 ), int((cy + cyt) //2 - 25 ))
+          midpoint2 = (int((cx + cxt) //2), int((cy + cyt) //2 + 50 ))
+          midpoint3 = (int((cx + cxt) //2), int((cy + cyt) //2 - 100 ))
```
AC408, line 70: what are these used for?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+          cv2.putText(image, f"{horiz_dist: .2f} in", midpoint, cv2.FONT_HERSHEY_COMPLEX, 1, (0, 101, 255), 2)
+          angle = np.degrees(np.arcsin(horiz_dist/(depth)))
```
AC408, line 72: what's angle?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+          cv2.putText(image, f"{angle:.2f} degrees", midpoint2, cv2.FONT_HERSHEY_COMPLEX, 1, (0, 101, 255), 2)
+          cv2.putText(image, f"{depth:.2f} in (depth)", midpoint3, cv2.FONT_HERSHEY_COMPLEX, 1, (255,0, 255), 2)
+          #Step 1 starts here
+          if idDict.get(markerId) == "R2": ##TODO: Add B2, L2, F2 in the future
+            #Step 1 is over. Start Step 2: move forward until tag out of sight. "
```
AC408, line 77: it sounds like this should be in a while loop?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+            if angle <= angle_zero:
+              motor.go_forward()
```
AC408, line 79: I think this might need a time delay, otherwise the robot goes forward and moves on (where the motor command can be overwritten or not reset)

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+def distance_to_camera(knownWidth, focalLength, perWidth):
+  # compute and return the distance from the maker to the camera
+  return (knownWidth * focalLength) / perWidth
+
+def direction_of_tag(corner,image):
+  (_, _, _, _) = corner
+  _,w = image.shape[0], image.shape[1]
+  cx, _ = int(960),int(540)
+  if  bottomLeft[0] > cx:
+    return "right"
```
AC408, line 24: nit, enums might be better

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+              motor.go_forward()
+              step1_done = True
+            else:
+              # Step 1 - "Keep rotating till we reach center of middle tag"
+              if direction_of_tag(corner[0],image) == "right":
+                motor.turn_left()
+              else: 
+                motor.turn_right()
+          else:
+            #All of this is Step 1
```
AC408, line 88: step 1 or step 2?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,127 @@
...
+              for i in range(num_calls):
+                motor.turn_right()
+            else: 
+              for i in range(num_calls):
+                motor.turn_left()
+            step3_done = True
+          
+          #step 4: reverse by 30 inches 
+          if step3_done: 
+            if (depth < 30):
```
AC408, line 110: I think this may need a while loop or delay

AC408 (changes requested):
made some comments on docking_script

april_tag_updates/docking_script.py
Shreya1233: it's both - they are interweaved

april_tag_updates/docking_script.py
Shreya1233: We know that the camera has 1080p resolution which means we already know the image has 1920 pixels horizontally by 1080 vertically. The center of the image is dividing these numbers by 2. It's given info that we can add to documentation on this file.

april_tag_updates/docking_script.py
Shreya1233: found experimentally!

april_tag_updates/docking_script.py
Shreya1233: Initially it was step 2 but now we just changed it to be part of step 1.

april_tag_updates/docking_script.py
Shreya1233: angle_zero is in degrees. Our camera is not accurate enough to actually be perfectly aligned with the center of the tag aka when the angle would be 0 degrees. angle_zero is 10 since we'll say if angle is 10, then the center of the camera is perfectly aligned with the center of the tag.

april_tag_updates/docking_script.py
Shreya1233: angle is the angle between the center of the camera to the center of the April tag!

april_tag_updates/docking_script.py
Shreya1233: it doesn't have to be because we are in a while true loop :) We will always continuously be checking this condition

Shreya1233: it doesn't have to be because we are in a while true loop :) We will always continuously be checking this condition

april_tag_updates/docking_script.py
AC408: yeah, I think it'll make sense to add to documentation

april_tag_updates/docking_script.py
AC408: maybe comment

april_tag_updates/docking_script.py
AC408: maybe comment?

april_tag_updates/docking_script.py
AC408: Oh, I see. Will this conflict with code for B2, L2, and F2 when they are added?

Shreya1233: it shouldn't! the reason we are adding it later is we first want to physically test one side aka the right side of the base station or tags R1, R2, R3 before we add B2, L2, and F2 to our code.

april_tag_updates/docking_script.py
AC408: ok. where's step 2?

april_tag_updates/docking_script.py
Shreya1233: Also I renamed this to angle_min to be more intuitive.

april_tag_updates/docking_script.py
Shreya1233: In step 1 we align with the center of the tag and move forward until the tag is out of sight and this all takes place under len(corners) > 0. Then we jump into step 2 which should start when we see no April tags(which is in the if step1_done and not step2_done condition which is under the if(Len(corners)<0) condition) . We reverse until we see an April tag which is when if len(corner) > 0. When an April tag is spotted, we set step2_done to true (we say we do this only if step1_done is true because we don't want to set step2_done to true where we just start our algorithm and we haven't even done step1.

april_tag_updates/docking_script.py
Shreya1233: if you to the top of the script, look under if len(corners) < 0 and under that look under if step1_done and not step2_done, step 2 starts there. Step 2 ends under if len(corners) > 0 and if step1_done: then step2_done.

april_tag_updates/docking_script.py
AC408: gotcha

april_tag_updates/docking_script.py
AC408: I see.It might be helpful to remove the comment on step 1 ending here then?

april_tag_updates/docking_script.py
AC408: sounds good! a comment might be nice as well

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,132 @@
...
+          ratio= known_width/pixel_width
+          dpix = np.sqrt((cx-cxt)**2+(cy-cyt)**2)
+          horiz_dist = dpix * ratio
+          midpoint = (int((cx + cxt) //2 ), int((cy + cyt) //2 - 25 ))
+          midpoint2 = (int((cx + cxt) //2), int((cy + cyt) //2 + 50 ))
+          midpoint3 = (int((cx + cxt) //2), int((cy + cyt) //2 - 100 ))
+          cv2.putText(image, f"{horiz_dist: .2f} in", midpoint, cv2.FONT_HERSHEY_COMPLEX, 1, (0, 101, 255), 2)
+          angle = np.degrees(np.arcsin(horiz_dist/(depth)))
+          cv2.putText(image, f"{angle:.2f} degrees", midpoint2, cv2.FONT_HERSHEY_COMPLEX, 1, (0, 101, 255), 2)
+          cv2.putText(image, f"{depth:.2f} in (depth)", midpoint3, cv2.FONT_HERSHEY_COMPLEX, 1, (255,0, 255), 2)
```
sazhang02, line 78: Most of this code seems to be for graphing/visuals on the camera ide popup screen thingy. Let's abstract this into functions like plot_line_between_tags and so. If we need to set angle and depth we could return these values from the function and after the func call in the loop.

april_tag_updates/docking_script.py
sazhang02: ^^ pls comment

april_tag_updates/docking_script.py
sazhang02: ^ comment for this would be g8

april_tag_updates/docking_script.py
sazhang02: For cx, cy:
cx, _ was initialized globally. Should add cy instead of _ and so u have `cx, cy = int(960),int(540)` done once

also functions for these as mentioned in other comment

sazhang02 (changes requested):
Hmm I'm getting a bit tripped up with following the logic. High-level what we want to do in each step makes a lot of sense and is clear in the google doc. However, I think one reason why it's confusing is bc the actual **implementation** of the high-level process is different. This is fine (and expected since its the implementation) but also a bit confusing since explanations like [responses to comments](https://github.com/cornellnexus/src/pull/120/files#r1141094888) are **not documented in the code** and **the code implementation differs from the general high-level explanation** of what the robot is doing + there are a bunch of conditionals/checks. I also think it's difficult to document explanations like the above in the code bc there are so many nested ifs so it gets spread out across many lines + interweaved w/ other logic.

**Along with adding some more comments/explanations** in code to describe how we're implementing the high-level process (e.g. aligning robot w middle tag), here's a few ways that could make this easier to read/understand:

**1) Assign conditions to variables or function outputs**
For example, if len(corner) > 0 is a condition we check more than once but I was very confsued about what this conditional meant till I read your explanation.

Since the tag computation for corners occurs once at the beginning of the while loop (`(corners, ids, rejected) = cv2.aruco.detectMarkers(image, arucoDict, parameters=arucoParams)`) , you could put initialize the condition result right below it as `is_april_tag_in_view = len(corner) > 0`. However, based on the logic in the google doc, it seems to me like we **should** be recomputing the corners before this check?? If you do need it multiple times and needs to be recalculated each time, you could abstract these two lines into a function and whenever you have the condition len(corner) > 0 call the function.

**2) Avoid nesting if statements**
There are **a lot** of if statements in this script. For example, 
Nested ifs under the condition `if len(corners) == 0`
```
    if len(corners) == 0:
      if angle < angle_min and depth < depth_min:
        step1_done  = True
        ...
```
and on the same level, a huge about of nested ifs under the condition `if len(corners) > 0`
```
    if len(corners) > 0:
      if step1_done:
        step2_done = True
       ....
```
if this if statement isn't passed then we basically "expect" a huge chunk of lines to just "not run" (hence the presence of no else). In general when you're nesting a bunch of if conditions for logic it becomes harder to read and understand unless you are already aware of what should happen. 

In these cases it's good to use a guard clause to tell the user they should only continue if a given set of logic is true. I don't think python has guard clauses tho, but this [page](https://sourcery.ai/blog/explaining-refactorings-3/#:~:text=Add%20guard%20clause&text=This%20can%20be%20even%20more,convert%20conditionals%20into%20guard%20clauses.&text=We%20add%20a%20guard%20clause%20by%20inverting%20the%20condition%20and%20returning%20immediately.) explains the issue with huge nested if statements and a python alternative (i.e. negating condition and returning sooner or in this case breaking out of the loop sooner). This prob explains it better than me so pls take a look. It's also good to do use explicit naming for conditions (aka suggestion 1) to make the logic easier to understand when negated/broken out early.

**3) Use enums for steps rather than flags**
Another suggestion is to use enums for the different steps 0-5 and a single variable for your step. From what I'm understanding you can only be in one step at a given time? You could make the assumption that if ur in step 2, you are not in step 1 (e.g. whenever step1_done was set to true just make ur step variable step2). Imo it's easier to read lines like `if <step1 done condtion> step = Step.Step2` and check `if step == Step.Step2`.
I would think for example lines `if step1_done and not step2_done` reduces down to just `if step == Step.Step2` and `if step3_done and not step4_done` reduces down to just `if step == Step.Step3`.

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,132 @@
...
+import time 
+from enum import Enum
+
+arucoDict = cv2.aruco.Dictionary_get(cv2.aruco.DICT_APRILTAG_36h11)
+arucoParams = cv2.aruco.DetectorParameters_create()
+cap = cv2.VideoCapture(0)
+window = 'Camera'
+cv2.namedWindow(window)
+def distance_to_camera(knownWidth, focalLength, perWidth):
+  # compute and return the distance from the maker to the camera
```
juliehelenek, line 16: is this supposed to say marker

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,132 @@
...
+  return (knownWidth * focalLength) / perWidth
+
+def direction_of_tag(corner,image):
+  (_, _, _, _) = corner
```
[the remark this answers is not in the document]
juliehelenek: agree

april_tag_updates/docking_script.py
juliehelenek: because theres no documentation, i dont know what cx is and i dont really understand the relation that this has to bottomLeft and determining right or left.

does right/left signify that we need to turn right/turn left?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,132 @@
...
+  cx, _ = int(960),int(540)
+  if  bottomLeft[0] > cx:
+    return "right"
+  else:
+    return "left"
+
+idDict = {1: "R1", 2: "R2", 3: "R3"}#TODO add all the other tags
+visited = []
+timeout =  10
+angle_min = 10
```
juliehelenek, line 31: degrees or radians?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,132 @@
...
+depth_min = 8
```
juliehelenek, line 32: unit of this?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,132 @@
...
+angle = float("inf")
+depth = float("inf")
+while True:
+    if start_time is None:
+        start_time = time.time()
+    ret, image = cap.read()
+    (corners, ids, rejected) = cv2.aruco.detectMarkers(image, arucoDict, parameters=arucoParams)
+    if len(corners) == 0:
+      if angle < angle_min and depth < depth_min:
+        step1_done  = True
```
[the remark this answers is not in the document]
juliehelenek: ^^^^

april_tag_updates/docking_script.py
juliehelenek: what does rn mean?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,132 @@
...
+      if time.time() - start_time > timeout and step1_done is False:
+        #Step 1 - Check for april tags rn by rotating in circle for 15 seconds.
+          motor.turn_right()
```
juliehelenek, line 51: i think eventually we're going to want to replace this with spin_motors?

juliehelenek (changes requested):

april_tag_updates/docking_script.py
Shreya1233: yes that's correct! It's our future to-do.

april_tag_updates/docking_script.py
Shreya1233: @juliehelenek I forgot to delete that it means right now

Shreya1233: @sazhang02 Yeah if no tags are found after 15 seconds, we go back to the return phase since we aren't close enough for docking phase to execute. Also I think having time as a parameter would be great since it'll reduce the number of lines we have by a alot. I can talk to them about it tmrw.

april_tag_updates/docking_script.py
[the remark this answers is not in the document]
Shreya1233: This is checking to see if we have the required min angle and distance to start executing step 2.

april_tag_updates/docking_script.py
Shreya1233: You use enums when the meaning of a variable is ambiguous. its not here. I think we good :))))

april_tag_updates/docking_script.py
Shreya1233: done!

april_tag_updates/find_distance_to_tag.py
```diff
@@ -0,0 +1,93 @@
+expected_dist = 35+2
+known_width = 4.18 ##inches
+focal_length = 1.37478492e+03
```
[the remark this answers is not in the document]
Shreya1233: pixels

april_tag_updates/docking_script.py
Shreya1233: degrees

april_tag_updates/docking_script.py
Shreya1233: visited is a list of tags that we visited already, timeout is in seconds, angle_min is in degrees, and depth_min is in inches.

april_tag_updates/new_april.py
```diff
@@ -0,0 +1,151 @@
...
+	"DICT_7X7_50": cv2.aruco.DICT_7X7_50,
+	"DICT_7X7_100": cv2.aruco.DICT_7X7_100,
+	"DICT_7X7_250": cv2.aruco.DICT_7X7_250,
+	"DICT_7X7_1000": cv2.aruco.DICT_7X7_1000,
+	"DICT_ARUCO_ORIGINAL": cv2.aruco.DICT_ARUCO_ORIGINAL,
+	"DICT_APRILTAG_16h5": cv2.aruco.DICT_APRILTAG_16h5,
+	"DICT_APRILTAG_25h9": cv2.aruco.DICT_APRILTAG_25h9,
+	"DICT_APRILTAG_36h10": cv2.aruco.DICT_APRILTAG_36h10,
+	"DICT_APRILTAG_36h11": cv2.aruco.DICT_APRILTAG_36h11
+}
```
jinlin270, line 25: It looks like we have ARUCO_DICT in 3 files. If they are the same, would it be a good idea to move it to a separate file and import it where it is needed?

april_tag_updates/new_april.py
```diff
@@ -0,0 +1,151 @@
...
+while True:
+    Dist = []
+    ret, image = cap.read()
+    image = imutils.resize(image, width=800)
+    (corners, ids, rejected) = cv2.aruco.detectMarkers(
+        image, arucoDict, parameters=arucoParams)
+    if len(corners) <= 0:
+        if CACHED_PTS is not None:
+            corners = CACHED_PTS
+    if len(corners) > 0:
```
jinlin270, line 52: can change to else, with comment len(corners) > 0.

april_tag_updates/new_april.py
```diff
@@ -0,0 +1,151 @@
...
+        CACHED_PTS = corners
+        if ids is not None:
+           ids = ids.flatten()
+           CACHED_IDS = ids
+        else:
+           if CACHED_IDS is not None:
```
jinlin270, line 58: for each branch, can we add the logic of what we are checking and why it matters?

april_tag_updates/new_april.py
```diff
@@ -0,0 +1,151 @@
...
+            cv2.putText(image2, str(
+                int(markerId)), (int(topLeft[0]-10), int(topLeft[1]-10)), cv2.FONT_HERSHEY_COMPLEX, 1, (0, 0, 255))
+            Dist.append((cX, cY))
+            # print(arucoDict)
+            if len(Dist2) == 0:
+                if Line_Pts2 is not None:
+                    Dist2 = Line_Pts2
+            if len(Dist2) == 2:
+                Line_Pts2 = Dist2
+            if len(Dist2) == 2:
```
jinlin270, line 140: these are checking the same conditions

april_tag_updates/new_april.py
```diff
@@ -0,0 +1,151 @@
...
+            bottomLeftPoint = (int(bottomLeft[0]), int(bottomLeft[1]))
+            cv2.line(image2, topLeftPoint, topRightPoint, (0, 255, 0), 2)
+            cv2.line(image2, topRightPoint, bottomRightPoint, (0, 255, 0), 2)
+            cv2.line(image2, bottomRightPoint, bottomLeftPoint, (0, 255, 0), 2)
+            cv2.line(image2, bottomLeftPoint, topLeftPoint, (0, 255, 0), 2)
+            cX = int((topLeft[0] + bottomRight[0])//2)
+            cY = int((topLeft[1] + bottomRight[1])//2)
+            measure2 = abs(5/(topLeft[0]-cX))
+            cv2.circle(image2, (cX, cY), 4, (255, 0, 0), -1)
+            cv2.putText(image2, str(
```
jinlin270, line 131: the logic for image 2 looks exceedingly similar to image 1. Can we instead make a function and call it on image 1 and image2?

jinlin270 (changes requested):

april_tag_updates/find_distance_to_tag.py
```diff
@@ -0,0 +1,93 @@
...
+ 					midpoint3 = (int((cx + cxt) //2), int((cy + cyt) //2 - 100 ))
+ 					cv2.putText(image, f"{horiz_dist: .2f} in", midpoint, cv2.FONT_HERSHEY_COMPLEX, 1, (0, 101, 255), 2)
+ 					# print(d_in)
+ 					angle = np.degrees(np.arcsin(horiz_dist/(vert_distance)))
+ 					cv2.putText(image, f"{angle:.2f} degrees", midpoint2, cv2.FONT_HERSHEY_COMPLEX, 1, (0, 101, 255), 2)
+ 					cv2.putText(image, f"{vert_distance:.2f} in (vert)", midpoint3, cv2.FONT_HERSHEY_COMPLEX, 1, (255,0, 255), 2)
+ 					# print(f"{angle} degrees")
+ 		cv2.imshow("yolo", image)
+ 		key = cv2.waitKey(1) & 0xFF
+ 		if key == ord('q'):
```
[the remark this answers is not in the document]
Shreya1233: right now at least, the user is the one who terminates the script otherwise it would go on forever. when we incorporate it in our code, the robot successfully docking at the base station should terminate the script.

april_tag_updates/docking_script.py
jinlin270: What are midpoint, midpoint2, and midpoint3? Documentation 👉👈?

april_tag_updates/docking_script.py
```diff
@@ -0,0 +1,178 @@
...
+motor: motor controller with commands to physically move the robot
+start_time: the timestamp when script execution began
+angle: current angle between center of camera and center of april tag
+depth: distance between center of camera and center of april tag
+timeout: experimentally determined 10-second duration after which, if the AprilTag is not detected, the rotation process begins
+angle_min: the minimum, physically possible angle between center of camera and center of april tag
+depth_min: the minimum, physically possible distance between center of camera and center of april tag
+arucoDict: A predefined ArUco dictionary containing the AprilTag 36h11 family.
+arucoParams: parameters that define various settings and thresholds used during the marker detection process.
+"""
```
sazhang02, line 33: Centralize specification
```suggestion
    """
    An enumeration of the different steps in the April Tag Docking Algorithm
    
    Arguments:
        cx,cy: center point of an image captured by a camera with 1080p resolution 
        idDict: dictionary of ids of April Tags mapped to corresponding string values. 
        visited: set of non-centered april tags that have been already been seen and aligned with camera. 
        step: the current step the docking algorithm is on  
        motor: motor controller with commands to physically move the robot
        start_time: the timestamp when script execution began
        angle: current angle between center of camera and center of april tag
        depth: distance between center of camera and center of april tag
        timeout: experimentally determined 10-second duration after which, if the AprilTag is not detected, the rotation process begins
        angle_min: the minimum, physically possible angle between center of camera and center of april tag
        depth_min: the minimum, physically possible distance between center of camera and center of april tag
        arucoDict: A predefined ArUco dictionary containing the AprilTag 36h11 family.
        arucoParams: parameters that define various settings and thresholds used during the marker detection process.
    """
    Step1 = 1
    Step2 = 2
    Step3 = 3
    Step4 = 4
    Done = 5
```

april_tag_updates/docking_script.py
sazhang02: hmm i think saying enums convey meanings when ambiguous is only one aspect of the benefits for enums.

Imo enums are labels for categories. Strings can be anything Let's say someone uses another func that returns "Left" instead of "left", then your equality check is invalid. Or someone can return "down" or change your code in the future thinking it's another direction since strings can be anything. Enums make the possibilities finite, explicitly conveyed, and error prone + compiler checking benefits (e.g. switch cases, check if valid direction).

sazhang02 (changes requested):
lgtm! just concerned about the directional strings

sazhang02 (changes requested):
lgtm! just concerned about the directional strings

april_tag_updates/camera_calibration.py
```diff
@@ -0,0 +1,67 @@
...
+ret, mtx, dist, rvecs, tvecs = cv2.calibrateCamera(objpoints, imgpoints, gray.shape[::-1], None, None)
+
+print("Camera matrix : \n")
+print(mtx)
+print("dist : \n")
+print(dist)
+print("rvecs : \n")
+print(rvecs)
+print("tvecs : \n")
+print(tvecs)
```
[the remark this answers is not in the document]
Shreya1233: oh we just ran this script once and we've never run it again. so I think it's fine to keep it as is. We only run this script every time we get a new camera to get the focal length of the camera!

PR #121, opened 2023-03-23
Turning Nodes Grid 

jinlin270:
## Summary
This PR removes intermediate nodes in the middle of traversing a straight line so the PID will be able to smoothly traverse a straight line without stopping and starting at each node in between. 

### Description

This PR removes intermediate nodes in the middle of traversing a straight line and only keeps nodes at the ends where the robot is turning (and the nodes during the circular turning for better traversal for visualization). This change is so that during traversal, the PID will be able to smoothly traverse a straight line instead of stopping and starting at each node in between. This change should not affect how the traversal work, so everything should look the same. 

## Test Plan

The test plan is to test the rectangle, triangle, and circle with is_vertical = True/ False in test_activation_shapes.py and check the outputted traversal graph with the graphs in tests/functionality_tests/activate_shape_graphs. The graphs should be the same as before.
Also, there is a print statement in grid.py for print(self.waypoints)，which should show that the nodes in between straight lines are not in the waypoints. 

## Other

Currently, for plotting and visualization purposes, the nodes during turning are still in the self.waypoints. However, these nodes should be removed later and replaced by a turn function, which should make the robot turn from the current waypoint to the goal waypoint.

### DIFF
--- a/engine/grid.py
+++ b/engine/grid.py
@@ -576,8 +576,9 @@ class Grid:
             next_row = self.curr_pos[1] + 1
 
         # Traverse the row until it is time to turn
+
         while not self.is_before_turning_column(turning_column):
-            self.waypoints.append(new_pos)
+            # self.waypoints.append(new_pos)
             self.curr_pos = new_pos
             new_pos = self.get_next_traversal_pos()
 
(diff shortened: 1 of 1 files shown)


AdityaKompella6 (approved):
Good Job

PR #123, opened 2023-03-26
Break beam new code with timer and full functionality

VDhriti:
## References
The older breakbeam pr

### Description 
Code expanded to take into account 4 sensors + do all the work the sensor was supposed to do: including differentiating stuff falling in the bucket from the bucket actually being full, and then returning the 50% full vs 100% full message depending on which sensors give true positive readings of the beam being broken.

## Test Plan 
Timer has been tested with two sensors, we are yet to test the whole class with 4 sensors and the 50% vs 100% function.

## Other
Worked with Amber (electrical) and Andrew (software)

### DIFF
--- a/electrical/breakbeam.py
+++ b/electrical/breakbeam.py
@@ -2,14 +2,14 @@ import RPi.GPIO as GPIO
 import time
 from enum import Enum
 
+
 class Breakbeam:
-    ACCEPTABLE_TIME = 50 # time to wait to confirm a full break
+    ACCEPTABLE_TIME = 3 # time to wait to confirm a full break
     class BEAM_PINS(Enum):
         HALF1 = 17
         HALF2 = 22
         FULL1 = 23
         FULL2 = 24
-    
 
     def __init__(self):
         """
@@ -18,9 +18,8 @@ class Breakbeam:
         """
         for BEAM_PIN in Breakbeam.BEAM_PINS:
             GPIO.setmode(GPIO.BCM)
-            GPIO.setup(BEAM_PIN, GPIO.IN, pull_up_down=GPIO.PUD_UP)
-            GPIO.add_event_detect(BEAM_PIN, GPIO.BOTH, callback=self.break_beam_callback)
-        self.beam_broken = False
+            GPIO.setup(BEAM_PIN.value, GPIO.IN, pull_up_down=GPIO.PUD_UP)
+            GPIO.add_event_detect(BEAM_PIN.value, GPIO.BOTH, callback=self.break_beam_callback)
         self.blocked_sensors = set()
     
     def break_beam_callback(self, channel):
@@ -28,14 +27,12 @@ class Breakbeam:
         Returns True if the beam at channel has been broken.
         Parameters: channel - the pin used by this sensor
         """
-        self.beam_broken = not GPIO.input(channel)
-        if self.beam_broken:
+        if not GPIO.input(channel):
             self.timer(channel)
             self.check_half()
             self.check_full()
-        # print("50%: " + self.check_half())
-        # print("100%: " + self.check_full())
-        return self.beam_broken
+        # print(self.check_half())
+        # print(self.check_full())
     
     def timer(self, channel):
         """
@@ -48,11 +45,23 @@ class Breakbeam:
         """
         start = time.time()
         # loops until ACCEPTABLE_TIME and then goes into the else statement once
-        while time.time() - start > Breakbeam.ACCEPTABLE_TIME:
-            pass
+        while time.time() - start < Breakbeam.ACCEPTABLE_TIME:
+            if(GPIO.input(channel)):
+                break
         else:
             self.blocked_sensors.add(Breakbeam.BEAM_PINS(channel).name)
-            # return channel (for testing)
+            self.blocked_sensors_check()
+            print(self.blocked_sensors)
+        
+    def blocked_sensors_check(self):
+        TO_BE_REMOVED = []
+        for sensor in self.blocked_sensors:
+            if(GPIO.input(Breakbeam.BEAM_PINS[sensor].value)):
+                TO_BE_REMOVED.append(sensor)
+                
+        for sensor in TO_BE_REMOVED:
+            self.blocked_sensors.remove(sensor) 
+        
         
     def check_half(self):
         """
@@ -76,5 +85,6 @@ class Breakbeam:
         
 
 # Allows to exit program and clean up Pi for testing
+breakbeam = Breakbeam()
 message = input("Press enter to quit\n\n")
 GPIO.cleanup()
--- a/electrical/breakbeam.py
+++ b/electrical/breakbeam.py
@@ -15,6 +15,7 @@ class Breakbeam:
         """
         Initializes each sensor to detect changes. Initializes beam_broken to False.
         Creates an empty list for the beams that are broken.
+        Attribute blocked_sensors: set of sensors that have been fully blocked.
         """
         for BEAM_PIN in Breakbeam.BEAM_PINS:
             GPIO.setmode(GPIO.BCM)
@@ -31,8 +32,8 @@ class Breakbeam:
             self.timer(channel)
             self.check_half()
             self.check_full()
-        # print(self.check_half())
-        # print(self.check_full())
+        # print("50% full: " + str(self.check_half()))
+        # print("100% full: " + str(self.check_full()))
     
     def timer(self, channel):
         """
@@ -51,10 +52,15 @@ class Breakbeam:
         else:
             self.blocked_sensors.add(Breakbeam.BEAM_PINS(channel).name)
             self.blocked_sensors_check()
-            print(self.blocked_sensors)
+            # print(self.blocked_sensors)
         
     def blocked_sensors_check(self):
-        TO_BE_REMOVED = []
+        """
+        Checks whether the sensors in the set blocked_sensors are still blocked.
+        Uses list TO_BE_REMOVED to store the sensors that have been unblocked and
+        then remove them from the set of blocked sensors.
+        """
+        TO_BE_REMOVED = [] # sensors to remove due to being unbroken
         for sensor in self.blocked_sensors:
             if(GPIO.input(Breakbeam.BEAM_PINS[sensor].value)):
                 TO_BE_REMOVED.append(sensor)
@@ -67,21 +73,13 @@ class Breakbeam:
         """
         Returns whether the 50% sensors are in the set of fully blocked sensors
         """
-        if len(self.blocked_sensors) < 4 and Breakbeam.BEAM_PINS(17).name in self.blocked_sensors and Breakbeam.BEAM_PINS(22).name in self.blocked_sensors:
-            # bucket has been 50% filled
-            return True
-        else:
-            return False
+        return len(self.blocked_sensors) < 4 and Breakbeam.BEAM_PINS(17).name in self.blocked_sensors and Breakbeam.BEAM_PINS(22).name in self.blocked_sensors
         
     def check_full(self):
         """
         Returns whether the 100% sensors are in the set of fully blocked sensors.
         """
-        if len(self.blocked_sensors) == 4:
-            # bucket has been 100% filled
-            return True
-        else:
-            return False
+        return len(self.blocked_sensors) == 4
         
 
 # Allows to exit program and clean up Pi for testing
--- a/electrical/breakbeam.py
+++ b/electrical/breakbeam.py
@@ -37,11 +37,11 @@ class Breakbeam:
     
     def timer(self, channel):
         """
-        Returns the pin number of this sensor.
         Gets the current time and calculates the time that has passed, then compares
         this to the ACCEPTABLE_TIME to determine whether the broken beam was just
         tripped once or if it is fully broken. If fully broken, add it to the set
-        blocked_sensors.
+        blocked_sensors. If the sensor becomes unblocked during tis ACCEPTABLE_TIME,
+        break out of the while-else loop.
         Parameters: channel - the pin used by this sensor
         """
         start = time.time()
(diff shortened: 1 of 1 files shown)


electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
+import RPi.GPIO as GPIO
+import time
+
+class Breakbeam:
+    ACCEPTABLE_TIME = 50 # in seconds
```
Andrew-Lowitt, line 5: Why are we choosing 50 seconds to be acceptable - Seems a bit arbitrary? May not necessarily be a problem, however, until we have a better reference, lets keep any eye out.

sazhang02: ^^ if this is a temp/experimental value maybe we can make a note in comments

electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
...
+    BEAM_PINS = {17:'50%_1', 22:'50%_2', 23:'100%_1', 24:'100%_2'}
+
+    def __init__(self):
+        for BEAM_PIN in Breakbeam.BEAM_PINS.keys():
+            GPIO.setmode(GPIO.BCM)
+            GPIO.setup(BEAM_PIN, GPIO.IN, pull_up_down=GPIO.PUD_UP)
+            GPIO.add_event_detect(BEAM_PIN, GPIO.BOTH, callback=self.break_beam_callback)
+            print(self.timer(BEAM_PIN))
```
sazhang02, line 13:
```suggestion
           self.timer(BEAM_PIN)
```
do we need the print statement? event to print timer has its own print statements in the function

electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
...
+        self.beam_broken = False
+        self.blocked_sensors = []
+    
+    def break_beam_callback(self, channel):
+        print(not GPIO.input(channel), channel)
```
sazhang02, line 18:
```suggestion
```
Remove print?

electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
...
+        self.beam_broken = not GPIO.input(channel)
+        if not GPIO.input(channel):
+            self.timer(channel)
```
sazhang02, line 21:
```suggestion
        self.beam_broken = not GPIO.input(channel)
        if self.beam_broken:
            # Bream beak detected interruption, run timer check to see whether it is a false reading
            self.timer(channel)
```

electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
+import RPi.GPIO as GPIO
+import time
+
+class Breakbeam:
+    ACCEPTABLE_TIME = 50 # in seconds
+    BEAM_PINS = {17:'50%_1', 22:'50%_2', 23:'100%_1', 24:'100%_2'}
```
sazhang02, line 6: Let's make an enum to represent the different values rather than strings.

electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
...
+            self.blocked_sensors.append(Breakbeam.BEAM_PINS[channel])
+            self.check_full()
+            return channel
+        
+    def check_full(self):
+        if len(self.blocked_sensors) == 2 and '50%_1' in self.blocked_sensors and '50%_2' in self.blocked_sensors:
+            return 'it was the 50 layer'
+        
+        if len(self.blocked_sensors) == 4:
+            return 'its full'
```
sazhang02, line 43:
```suggestion
        if len(self.blocked_sensors) == 2 and '50%_1' in self.blocked_sensors and '50%_2' in self.blocked_sensors:
            # it was the 50 layer
            return False
        
        if len(self.blocked_sensors) == 4:
            # it's full
            return True
```
would be cleaner

But actually could we just make this function to return the bucket level? or make separate function that does this (need for gui x.x)

sazhang02 (changes requested):
Good job with this PR! Added some comments

electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
...
+    BEAM_PINS = {17:'50%_1', 22:'50%_2', 23:'100%_1', 24:'100%_2'}
+
+    def __init__(self):
+        for BEAM_PIN in Breakbeam.BEAM_PINS.keys():
+            GPIO.setmode(GPIO.BCM)
+            GPIO.setup(BEAM_PIN, GPIO.IN, pull_up_down=GPIO.PUD_UP)
+            GPIO.add_event_detect(BEAM_PIN, GPIO.BOTH, callback=self.break_beam_callback)
+            print(self.timer(BEAM_PIN))
+        self.beam_broken = False
+        self.blocked_sensors = []
```
sazhang02, line 15: also do we care about the order theyre blocked in? if not a set would be more efficient here + removes possibilities of duplicates being added

electrical/breakbeam.py
VDhriti: It is a bit arbitrary rn, couldn't think of a better reference yeah. I'll add a note. Should I make a new commit with the changes?

electrical/breakbeam.py
VDhriti: I don't think we do, dunno why we have that there whoops

electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
...
+            if GPIO.input(channel) == False:
+                print('false reading')
+                if Breakbeam.BEAM_PINS[channel] in self.blocked_sensors: #it doesnt log duplicate values
+                    self.blocked_sensors.remove(Breakbeam.BEAM_PINS[channel])
+                break
+        else:
+            print('Time passed was ', time.time() - start, 'seconds')
+            self.blocked_sensors.append(Breakbeam.BEAM_PINS[channel])
+            self.check_full()
+            return channel
```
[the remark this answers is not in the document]
sazhang02: update -- learned that while-else loops are a thing from Dhriti. only concern left is if we need self.beam_broken

sazhang02:
potential concern to keep in mind (don't need to address now imo) is the while loop in timer might halt the rest of the code from proceeded. not sure if gpio runs callback on a diff thread 🤔  we can check for this when we integrate this code with our current implementation w other sensors + code!

electrical/breakbeam.py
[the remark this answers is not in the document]
juliehelenek: """
    Return (What does the function return?)
    (One line description of what else the function does, if applicable)
    Parameters: Description of parameters
    Precondition: Description of precondition on any parameters
    """

electrical/breakbeam.py
juliehelenek: is there a minimum number of sensors that need to be blocked in order for us to deduce that we're 50% filled / 100% filled?

electrical/breakbeam.py
juliehelenek: ^^^^^^

juliehelenek: if you make a variable, try to use it as much as u can! (ie use self.beam_broken instead of writing not GPIO.input(channel) again)

electrical/breakbeam.py
```diff
@@ -0,0 +1,47 @@
...
+        self.blocked_sensors = []
+    
+    def break_beam_callback(self, channel):
+        print(not GPIO.input(channel), channel)
+        self.beam_broken = not GPIO.input(channel)
+        if not GPIO.input(channel):
+            self.timer(channel)
+        return self.beam_broken
+    
+    def timer(self, channel):
```
juliehelenek, line 24: specification for this function pls

electrical/breakbeam.py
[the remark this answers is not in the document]
juliehelenek: i've never used a while-else loop and i teach CS1110 LOL
can we add a comment above it to describe how it works (in a few words?)
for example;
#loops until ACCEPTABLE_TIME and then goes into the else statement once

juliehelenek (changes requested):
i agree with all of sammie's comments! address these and then ask for a re-review!

electrical/breakbeam.py
```diff
@@ -0,0 +1,90 @@
...
+    def __init__(self):
+        """
+        Initializes each sensor to detect changes. Initializes beam_broken to False.
+        Creates an empty list for the beams that are broken.
+        """
+        for BEAM_PIN in Breakbeam.BEAM_PINS:
+            GPIO.setmode(GPIO.BCM)
+            GPIO.setup(BEAM_PIN.value, GPIO.IN, pull_up_down=GPIO.PUD_UP)
+            GPIO.add_event_detect(BEAM_PIN.value, GPIO.BOTH, callback=self.break_beam_callback)
+        self.blocked_sensors = set()
```
juliehelenek, line 23: blocked_sensors is an attribute for your Breakbeam class - let's add this into the specification for _ _ init _ _ function

electrical/breakbeam.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        start = time.time()
+        # loops until ACCEPTABLE_TIME and then goes into the else statement once
+        while time.time() - start < Breakbeam.ACCEPTABLE_TIME:
+            if(GPIO.input(channel)):
+                break
+        else:
+            self.blocked_sensors.add(Breakbeam.BEAM_PINS(channel).name)
+            self.blocked_sensors_check()
+            print(self.blocked_sensors)
```
juliehelenek, line 54: do we want to keep this print statement?

electrical/breakbeam.py
```diff
@@ -0,0 +1,90 @@
...
+        
+    def blocked_sensors_check(self):
```
juliehelenek, line 56: can we add specification to this function? what does TO_BE_REMOVED mean? why do we even need it? can't we just do 

```
for sensor in self.blocked_sensors:
     if (GPIO.input(Breakbeam.BEAM_PINS[sensor].value)):
         self.blocked_sensors.remove(sensor)

File: electrical/breakbeam.py [hunk 13]
```diff
@@ -0,0 +1,90 @@
...
+            # bucket has been 50% filled
+            return True
+        else:
+            return False
+        
+    def check_full(self):
+        """
+        Returns whether the 100% sensors are in the set of fully blocked sensors.
+        """
+        if len(self.blocked_sensors) == 4:
```
juliehelenek, line 80: you can simplify this to 
return len(self.blocked_sensors) == 4

electrical/breakbeam.py
```diff
@@ -0,0 +1,90 @@
...
+                
+        for sensor in TO_BE_REMOVED:
+            self.blocked_sensors.remove(sensor) 
+        
+        
+    def check_half(self):
+        """
+        Returns whether the 50% sensors are in the set of fully blocked sensors
+        """
+        if len(self.blocked_sensors) < 4 and Breakbeam.BEAM_PINS(17).name in self.blocked_sensors and Breakbeam.BEAM_PINS(22).name in self.blocked_sensors:
```
juliehelenek, line 70: you can simplify this

return len(self.blocked_sensors) < 4 and Breakbeam.BEAM_PINS(17).name in self.blocked_sensors and Breakbeam.BEAM_PINS(22).name in self.blocked_sensors

electrical/breakbeam.py
```diff
@@ -0,0 +1,90 @@
...
+        Gets the current time and calculates the time that has passed, then compares
+        this to the ACCEPTABLE_TIME to determine whether the broken beam was just
+        tripped once or if it is fully broken. If fully broken, add it to the set
+        blocked_sensors.
+        Parameters: channel - the pin used by this sensor
+        """
+        start = time.time()
+        # loops until ACCEPTABLE_TIME and then goes into the else statement once
+        while time.time() - start < Breakbeam.ACCEPTABLE_TIME:
+            if(GPIO.input(channel)):
```
juliehelenek, line 49: what is this if statement checking? is this just an ece thing?

juliehelenek (commented):
added some comments to clean up your code a bit! good work!

electrical/breakbeam.py
AmberL235: We cannot change the size of the set while iterating over it, so instead we append to another set. Is there a better way to do this?

electrical/breakbeam.py
AmberL235: I believe it is an ece thing, where the GPIO.input(channel) retakes the reading when we call it but I'm not sure if that's how it works. We were discussing whether storing GPIO.input(channel) in a variable would make it store the reading for only that one time, so just to be safe we removed the variable and just use GPIO.input(channel).

electrical/breakbeam.py
AmberL235: Nope, it was just for testing and I guess we didn't comment that one out

electrical/breakbeam.py
AmberL235: But the if statement checks whether the sensor becomes unblocked and exits the while-else loop if this is the case.

electrical/breakbeam.py
sazhang02: try using the `pop` method?
https://stackoverflow.com/questions/31448465/how-to-loop-through-a-set-while-removing-items-from-the-set-in-python-3

electrical/breakbeam.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        Returns whether the 100% sensors are in the set of fully blocked sensors.
+        """
+        return len(self.blocked_sensors) == 4
+        
+
+# Allows to exit program and clean up Pi for testing
+breakbeam = Breakbeam()
+message = input("Press enter to quit\n\n")
+GPIO.cleanup()
```
sazhang02, line 88: Let's comment this out (otherwise when another file imports breakbeam it will run this script with it)

sazhang02 (approved):
Nice job with the changes!! Once ug address Julie's comments + the one about commenting out the last few lines should be good to merge this into main :) Going to approve!

electrical/breakbeam.py
AmberL235: I believe in python set pop() removes and returns a random item from the set, and doesn't take parameters to specify which to remove.
[https://www.w3schools.com/python/ref_set_pop.asp](url)

juliehelenek: yeah for lists, you can do list.pop(index) but probably doesnt work for sets - this should be fine then

juliehelenek (approved):
good work!!

PR #124, opened 2023-03-28
[Mission Refactoring 1] Abstract Control Mode into it's own file

sazhang02:
### Description

ControlMode should be in its own file as it is not specific to just the Mission class (also used in grid for example). This will make `Mission.py` cleaner. Since this enum is moved to a different file, the respective import statements are updated for the new pathing.

## Test Plan

Ran test cases using `python -m tests.compilation_tests.test_runner`

## Other

This is the first part of mission refactoring. Note that this is being merged into the branch `mission_refactor_feature`.

List of Mission Refactoring PR's:
[[Mission Refactoring 1] Abstract Control Mode into it's own file ](https://github.com/cornellnexus/src/pull/124)
[[Mission Refactoring 2] Centralize basestation calculations within Basestation Class initialization](https://github.com/cornellnexus/src/pull/125)
[[Mission Refactoring 3] Abstract attributes from Mission into a new class Mission_State](https://github.com/cornellnexus/src/pull/126)
[[Mission Refactoring 4] Centralize sensor attributes within Robot State, initialize in execute_setup](https://github.com/cornellnexus/src/pull/127)

PR #125, opened 2023-03-28
[Mission Refactoring 2] Centralize basestation calculations within Basestation Class initialization

sazhang02:
### Description

Right now, in `Mission.py`, the following is initialized:
```
        self.base_station_angle = base_station.heading
        self.allowed_docking_pos_error = allowed_docking_pos_error
        x = get_vincenty_x((grid.lat_min, grid.long_min),
                           base_station.position)
        y = get_vincenty_y((grid.lat_min, grid.long_min),
                           base_station.position)
        self.base_station_loc = (x, y)
```
Mission includes redundant attributes like `self.base_station_angle` which is just the heading of the base station parameter. Rather than creating new variables, it would be better to just save the base station instance as an attribute to the mission state. 

Additionally there is the calculation of the location of base station with respect to the Grid coordinate system (aka x,y then loc). I think it would make more sense to centralize the calculation of the base station location within the Base Station class rather than the Mission Class, and access its value through the base station attribute.

While going through the initializations, I noticed it was unused in tests/functionality_tests/obstacle_tracking_test.py and removed the import

## Test Plan

Ran `python -m tests.compilation_tests.test_runner`

## Other

This is the second part of mission refactoring. Note that this is being merged into the branch `controlmode_abstract` (the branch for the first PR https://github.com/cornellnexus/src/pull/124).
Note: Updating the initialization of Base Station is done in the third PR https://github.com/cornellnexus/src/pull/126 (since I commited it tg with those changes)

List of Mission Refactoring PR's:
[[Mission Refactoring 1] Abstract Control Mode into it's own file ](https://github.com/cornellnexus/src/pull/124)
[[Mission Refactoring 2] Centralize basestation calculations within Basestation Class initialization](https://github.com/cornellnexus/src/pull/125)
[[Mission Refactoring 3] Abstract attributes from Mission into a new class Mission_State](https://github.com/cornellnexus/src/pull/126)
[[Mission Refactoring 4] Centralize sensor attributes within Robot State, initialize in execute_setup](https://github.com/cornellnexus/src/pull/127)

### DIFF
--- a/engine/base_station.py
+++ b/engine/base_station.py
@@ -1,24 +1,38 @@
 import math
-
+from engine.kinematics import get_vincenty_x, get_vincenty_y
 
 class BaseStation:
     """
     Contains base-station-specific information.
     """
 
-    def __init__(self, position, heading=math.pi/2, battery=None, plastic_load=None):
+    def __init__(self, coord, grid, heading=math.pi/2, battery=None, plastic_load=None):
         """
         Arguments:
-            position: location of the base station in GPS coordinates in the form (latitude, longitude)
+            coord: location of the base station in GPS coordinates in the form (latitude, longitude)
+            grid: 
+            heading: which direction the base station is facing in terms of unit circle (in radians), by
+                default faces North (angle pi/2)
+            battery: TODO
+            plastic_load: TODO
+        Instance Attributes:
+            coord: location of the base station in GPS coordinates in the form (latitude, longitude)
+            position: location of base station in the Grid coordinate system
             heading: which direction the base station is facing in terms of unit circle (in radians), by
                 default faces North (angle pi/2)
             battery: TODO
             plastic_load: TODO
         """
-        self.position = position
+        self.coord = coord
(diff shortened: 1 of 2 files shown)


AC408 (approved):

jinlin270 (commented):
I am a bit confused on what the base_station stands for. Is it literally the stationary base_station, or is it the object storing the conditions necessary to determine if the robot should return to the base station?

Also, are the said redundant attributes in Mission.py removed/changed?

Aayush-Agnihotri (approved):

Shreya1233 (approved):

sazhang02:
@jinlin270 Hello! yes i think the BaseStation class is a representation of our literal base station on the beach (like where it is location). 

You bring up a good point about storing conditions like returning to the base station. Although this isn't a part of our base station class currently, I think this could be something for @Andrew-Lowitt to consider with the return to base station class.

also the redundant attributes are removed in https://github.com/cornellnexus/src/pull/126!

PR #126, opened 2023-03-28
[Mission Refactoring 3] Abstract attributes from Mission into a new class Mission_State

sazhang02:
### Description

The attributes of `Mission` are moved to `Mission State`. All references to these mission attributes in `Mission.py`and `sim_trajectory.py `are updated to reference the new attribute `mission_state`.

## Test Plan

Ran tests `python -m tests.compilation_tests.test_runner`
Ran `python -m engine.sim_trajectory` to make sure it compiled and nothing changed

## Other

This is the third part of mission refactoring. Note that this is being merged into the branch `basestation_init` (the branch for the second PR https://github.com/cornellnexus/src/pull/125).

List of Mission Refactoring PR's:
[[Mission Refactoring 1] Abstract Control Mode into it's own file ](https://github.com/cornellnexus/src/pull/124)
[[Mission Refactoring 2] Centralize basestation calculations within Basestation Class initialization](https://github.com/cornellnexus/src/pull/125)
[[Mission Refactoring 3] Abstract attributes from Mission into a new class Mission_State](https://github.com/cornellnexus/src/pull/126)
[[Mission Refactoring 4] Centralize sensor attributes within Robot State, initialize in execute_setup](https://github.com/cornellnexus/src/pull/127)

PR #127, opened 2023-03-28
[Mission Refactoring 4] Centralize sensor attributes within Robot State, initialize in execute_setup

sazhang02:
### Description

As mentioned in the TLDR ^

So, in order to do this we must remove sensor related attributes and initializations from `Mission` (and those being sent as arguments to `execute_setup`. Then, keep the sensor attributes in `Robot_State`. I decided to move the initialization of the sensor attributes from `Robot_State` to `execute_setup` since
1) The purpose of `execute_setup` is to setup our robot, sensors, and make sure the sensors are setup/working properly. So, I think it makes sense to initialize it + do any additional sensor setup all in one place.
2) There could be an interim period between when we want to initialize our robot and run our mission (whose first phase is setup). For example, we might not be ready to start the mission of traversing the beach, which requires all the sensors. The robot can exist without the sensors being initialized.

## Test Plan

This would need to be tested in more detail on the robot with all the sensors attached. 

Imo tho I think any sensor initializations in the robot class are be overridden anyways by `execute_setup` in the robot class which is called using the same sensor initializations in Mission, which is initialized after the robot anyways. So, based on this, if there are any issues from this they likely already existed in our current implementation.

Doesn't test these changes but I also,
Ran tests `python -m tests.compilation_tests.test_runner`
Ran `python -m engine.sim_trajectory` to make sure it compiled and nothing changed.

## Other

This is the fourth and final part of mission refactoring. Note that this is being merged into the branch `mission_state` (the branch for the third PR https://github.com/cornellnexus/src/pull/126).

List of Mission Refactoring PR's:
[[Mission Refactoring 1] Abstract Control Mode into it's own file ](https://github.com/cornellnexus/src/pull/124)
[[Mission Refactoring 2] Centralize basestation calculations within Basestation Class initialization](https://github.com/cornellnexus/src/pull/125)
[[Mission Refactoring 3] Abstract attributes from Mission into a new class Mission_State](https://github.com/cornellnexus/src/pull/126)
[[Mission Refactoring 4] Centralize sensor attributes within Robot State, initialize in execute_setup](https://github.com/cornellnexus/src/pull/127)

### DIFF
--- a/engine/mission.py
+++ b/engine/mission.py
@@ -17,7 +17,7 @@ class Mission:
         while self.mission_state.robot.robot_state.phase != Phase.COMPLETE:
             if self.mission_state.robot.robot_state.phase == Phase.SETUP:
                 # TODO: Determine if sensors are part of mission vs robot
-                self.mission_state.robot.execute_setup(self.mission_state.robot_radio_session, self.mission_state.gps, self.mission_state.imu, self.mission_state.motor_controller)
+                self.mission_state.robot.execute_setup()
 
             elif self.mission_state.robot.robot_state.phase == Phase.TRAVERSE:
                 self.mission_state.waypoints_to_visit = self.mission_state.robot.execute_traversal(self.mission_state.waypoints_to_visit,
--- a/engine/mission_state.py
+++ b/engine/mission_state.py
@@ -4,20 +4,6 @@ from engine.control_mode import ControlMode
 from engine.grid import Grid
 from engine.base_station import BaseStation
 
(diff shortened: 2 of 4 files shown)


jinlin270 (approved):
This looks good. Moving the sensor-related attributes and initialization into robot.py execute_setup makes a lot of sense, and the code looks much cleaner.

Aayush-Agnihotri (approved):

Shreya1233 (approved):

PR #128, opened 2023-04-15
Fixing obstacle avoidance algorithm with setting obstacle avoidance flag

AC408:
## Summary
Previously, the obstacle detection thread sets a flag to true if there is an obstacle and the mission state checks the flag and update the goal location and phase when the flag is true. Without this, the obstacle detection thread does not know the goal location and cannot update it appropriately. Now, we are changing this so that the goal location is automatically updated and the tracking obstacle thread can automatically change the phase.

### Description
Update goal_location and prev_phase for each fsm state that lead into obstacle avoidance. Added enable_obstacle_avoidance in the obstacle tracking thread to pause the tracking (for turning). Now that we can pause/resume the tracking obstacle thread, we can remove the roomba flags. Added a check for whether we are in obstacle avoidance for move to target node to interupt while loop.

## Test Plan
python -m tests.functionality_tests.obstacle_tracking_test
python -m tests.compilation_tests.test_runner
python -m engine.sim_trajectory

### DIFF
--- a/engine/robot.py
+++ b/engine/robot.py
@@ -131,7 +131,7 @@ class Robot:
         self.robot_state.goal_location = target
         self.loc_pid_x.reset_integral()
         self.loc_pid_y.reset_integral()
-        while distance_away > allowed_dist_error:
+        while (distance_away > allowed_dist_error) and not (self.robot_state.phase == Phase.AVOID_OBSTACLE):
             if self.robot_state.is_sim:
                 # Adding simulated noise to the robot's state based on gaussian distribution
                 self.robot_state.state[0] = np.random.normal(
--- a/engine/robot.py
+++ b/engine/robot.py
@@ -302,7 +302,6 @@ class Robot:
             self.move_to_target_node(
                 curr_waypoint, allowed_dist_error, database)
             unvisited_waypoints.popleft()
-            return unvisited_waypoints
             # TODO: THIS ISNT CORRECT: NEED TO CHECK IF AVOID_OBSTACLE IN move_to_target_node or can also make PID traversal a separate thread and stop the thread when obstacle detected
 
         self.set_phase(Phase.RETURN)
--- a/engine/robot.py
+++ b/engine/robot.py
@@ -276,12 +276,13 @@ class Robot:
         self.robot_state.prev_phase = Phase.TRAVERSE
         self.robot_state.control_mode = control_mode
         if control_mode == 4:  # Roomba mode
+            # roomba mode uses its own obstacle avoidance
+            self.robot_state.enable_obstacle_avoidance = False
             self.traverse_roomba(base_station_loc, time_limit, roomba_radius)
-            self.robot_state.is_roomba_traversal = True
+            self.robot_state.enable_obstacle_avoidance = True
         else:
             self.traverse_standard(unvisited_waypoints,
                                    allowed_dist_error, database)
-            self.robot_state.is_roomba_traversal = False
 
     def traverse_standard(self, unvisited_waypoints, allowed_dist_error, database):
         """ Move the robot by following the traversal path given by [unvisited_waypoints].
@@ -339,7 +340,7 @@ class Robot:
             # if moving will cause the robot to move through the obstacle
             is_next_timestep_blocked = next_radius < self.robot_state.detect_obstacle_range
             # sensor should not detect something in the robot
-            if (next_radius > roomba_radius) or (self.robot_state.is_roomba_obstacle and is_next_timestep_blocked):
+            if (next_radius > roomba_radius) or is_next_timestep_blocked:
                 # this needs to be synchronous/PID'ed, otherwise, turn might be called while robot moving forward
                 if self.robot_state.is_sim:
                     # for some reason I don't think this should work. This needs to be blocking: wait for the robot to finish going backward before turning
@@ -398,9 +399,7 @@ class Robot:
                     if curr_ultrasonic_value < self.robot_state.front_sensor_offset:
                         self.set_phase(Phase.FAULT)
                         return None
-                if self.robot_state.is_roomba_traversal:  # roomba mode
-                    self.robot_state.is_roomba_obstacle = True
-                elif (self.robot_state.phase == Phase.TRAVERSE) or (self.robot_state.phase == Phase.RETURN) or (self.robot_state.phase == Phase.DOCKING) or (
+                if (self.robot_state.phase == Phase.TRAVERSE) or (self.robot_state.phase == Phase.RETURN) or (self.robot_state.phase == Phase.DOCKING) or (
                         self.robot_state.phase == Phase.AVOID_OBSTACLE):
                     if curr_ultrasonic_value < self.robot_state.detect_obstacle_range:
                         # Note: didn't check whether we can reach goal before contacting obstacle because obstacle
--- a/engine/robot.py
+++ b/engine/robot.py
@@ -231,8 +231,6 @@ class Robot:
         print('heading: ' + str(self.robot_state.state[2]))
 
     def execute_setup(self):
-        self.robot_state.goal_location = None
-        self.robot_state.prev_phase = Phase.SETUP
         if not self.is_sim:
             self.robot_state.motor_controller = MotorController(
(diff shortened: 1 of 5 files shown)


engine/robot.py
```diff
@@ -227,19 +231,26 @@ def print_current_state(self):
         print('heading: ' + str(self.robot_state.state[2]))
 
     def execute_setup(self):
+        self.robot_state.goal_location = None
+        self.robot_state.prev_phase = Phase.SETUP
```
sazhang02, line 235: Can we move this initialization to the initialization of Robot_State?

engine/robot.py
```diff
@@ -127,9 +128,10 @@ def move_to_target_node(self, target, allowed_dist_error, database):
 
         # location error (in meters)
         distance_away = self.calculate_dist(target, predicted_state)
+        self.robot_state.goal_location = target
         self.loc_pid_x.reset_integral()
         self.loc_pid_y.reset_integral()
-        while distance_away > allowed_dist_error:
+        while (distance_away > allowed_dist_error) and not (self.robot_state.phase == Phase.AVOID_OBSTACLE):
```
sazhang02, line 134: Hmm just want to check my understanding here with the goal_location and our unvisited waypoints:

Lets say you were trying to move to target node X out of the waypoints [X, Y, Z]. You are still super far away from X but bam theres an obstacle in the way 🪨 . So the function breaks out of this while loop bc we don't want to traverse anymore we want to avoid obstacle (assuming the tracking function changes the phase here). However, at the conclusion of the `move_to_target_node` function we pop the leftmost unvisited waypoint (assuming the waypoint was fully visited if the function terminates), which is X but X was not fully visited. Ideally after we exit obstacle avoidance we would want to go back to moving to X. Can we assume visited X will be fully handled by obstacle avoidance by the time we return back to our traversal phase since we set the goal_location to that node aka `target`?

engine/robot.py
```diff
@@ -194,7 +196,7 @@ def turn_to_target_heading(self, target_heading, allowed_heading_error, database
             allowed_heading_error: the maximum error in radians a robot can have to target heading while turning in
                 place.
         """
-
+        self.robot_state.enable_obstacle_avoidance = False  # we dont want the robot to avoid obstacle here
```
sazhang02, line 199: smart

engine/robot.py
```diff
@@ -253,22 +264,25 @@ def execute_setup(self):
...
+                obstacle_avoidance = threading.Thread(
+                    target=self.track_obstacle, daemon=True)
                 obstacle_avoidance.start()  # spawn thread to monitor obstacles
                 self.set_phase(Phase.TRAVERSE)
         else:
             self.set_phase(Phase.TRAVERSE)
 
     def execute_traversal(self, unvisited_waypoints, allowed_dist_error, base_station_loc, control_mode, time_limit,
                           roomba_radius, database):
+        self.robot_state.prev_phase = Phase.TRAVERSE
```
sazhang02, line 276: shouldn't this be at the end of the function? bc we also have a attribute for the current phase and this seems to be considering the prev_phase as the current phase? maybe we can add `self.robot_state.prev_phase = self.robot_state.phase` in the `set_phase` function?

engine/robot.py
```diff
@@ -366,49 +377,48 @@ def track_obstacle(self):
...
-                if curr_ultrasonic_value < self.robot_state.detect_obstacle_range:
-                    # Note: didn't check whether we can reach goal before contacting obstacle because obstacle
-                    # detection does not detect angle, so obstacle could be calculated to be falsely farther away than
-                    # the goal. Not optimal because in cases, robot will execute boundary following when it can reach
-                    # goal
-                    self.set_phase(Phase.AVOID_OBSTACLE)
-                    if self.robot_state.is_sim:
-                        with open(ROOT_DIR + '/tests/functionality_tests/csv/avoid_obstacle_result.csv', 'a') as fd:
-                            fd.write("Avoid" + '\n')
+            if self.robot_state.enable_obstacle_avoidance:
```
sazhang02, line 380: Lmk what you think about this but since we are just nesting all the below code under this if statement i think it would be cleaner if separated out the opposing condition and handled it earlier to avoid nesting if statements. Thinking about the idea in 2) of the docking algorithm [here](https://github.com/cornellnexus/src/pull/120#pullrequestreview-1349568065) -- though the decision to guard here isn't quite as clean bc we still need the while True loop to keep looping (normally would just negate and return from the function early)

What do you think about this?

```
if not self.robot_state.enable_obstacle_avoidance:
    pass
if self.robot_state.is_sim:... [prev code]
```
Not sure which is better here bc i think in general reducing nested ifs when possible is cleaner but idk if the pass is weird.

engine/robot.py
```diff
@@ -574,12 +584,13 @@ def execute_return(self, base_loc, base_angle, allowed_docking_pos_error, allowe
             allowed_heading_error: the maximum error in radians a robot can have to target heading while turning in
                 place.
         """
-        docking_dist_to_base = 1.0  # how close the robot should come to base before starting DOCKING
+        self.robot_state.prev_phase = Phase.RETURN
```
sazhang02, line 587: Same as above

requirements.txt
```diff
@@ -8,7 +8,7 @@ vincenty==0.1.4
 statistics==1.0.3.5
 PySimpleGui == 4.59.0 #comment this out when running on rpi
 coverage
-numpy==1.21.6
+numpy==1.24.2
```
sazhang02, line 11: merge main \o/

sazhang02 (changes requested):
good job as always allen :D 👍  just a few questions

engine/robot.py
AC408: uhhh, seems a bit weird for me but I'm not opposed

AC408: also, I think it should be continue, not pass

engine/robot.py
AC408: I think we want it in the beginning so if obstacle avoidance triggers, phase becomes obstacle avoidance phase, but we want to go back to prev phase

engine/robot.py
AC408: I think we want it in the beginning so if obstacle avoidance triggers, phase becomes obstacle avoidance phase, but we want to go back to prev phase

engine/robot.py
```diff
@@ -253,22 +262,25 @@ def execute_setup(self):
...
         else:
             self.set_phase(Phase.TRAVERSE)
 
     def execute_traversal(self, unvisited_waypoints, allowed_dist_error, base_station_loc, control_mode, time_limit,
                           roomba_radius, database):
+        self.robot_state.prev_phase = Phase.TRAVERSE
         self.robot_state.control_mode = control_mode
         if control_mode == 4:  # Roomba mode
+            # roomba mode uses its own obstacle avoidance
+            self.robot_state.enable_obstacle_avoidance = False
```
jinlin270, line 278: is there a reason we set enable_obstacle_avoidance to False then to True? It doesn't look like this attribute is used in traverse_roomba.

AC408: this attribute is used in track_obstacle. We want to "pause" the obstacle avoidance thread bc roomba mode has its own obstacle avoidance

engine/robot.py
AdityaKompella6: This makes sense

engine/robot.py
AdityaKompella6: what does wheel radius equals 0 mean

AdityaKompella6 (approved):
This is good you are just kinda changing stuff to False so that's just and aesthetic change. Keep on making PR's my guy.

engine/robot.py
AC408: what?

Aayush-Agnihotri (approved):
Good work

jinlin270 (approved):
Looks good!

engine/robot.py
sazhang02: noice

engine/robot.py
sazhang02: whatever u think is best either fine!

engine/robot.py
sazhang02: yeppp

sazhang02 (approved):
YAYYY good job allen as always amazing WOOO :DDDDDD

PR #131, opened 2023-04-20
[RENAME sim_trajectory.py to main.py] Abstracting and cleaning up our main script

sazhang02:
## References

https://docs.python.org/3/howto/argparse.html

#### Problem
`sim_trajectory.py` was made with the intention of running a simulated version of our codebase. This is not good because we can't run this script on the robot (as it was MADE FOR a simulated environment) without making many code changes (e.g. comment all the matplotlib code out, change boolean variables for turning on transmission communication on/off, etc).  

#### Solution
So, this is problem addressed by 1) Abstracting all plotting code and 2) Moving all testing/development flags to a central `config.json` file and creating command-line input for testable arguments.

Additionally, the code in `sim_trajectory.py` was also just cleaned up a bunch bc there was outdated/decisions made w/ past knowledge (see full detail breakdown below in Changes).

Note: we only want one script to run our codebase (`main.py`) rather than two, one for a simulated environment(`sim_trajectory.py`) and one for the robot (`main.py`). Why? The point of the simulated environment script would be to test the script for the robot so at that point you would just be control-c control-ving the robot script code. So, it is better to just centralize our main script to run the codebase and add flags to differentiate and customize testing in a simulated environment, for more flexibility.

#### Changes
FIle breakdown
- `config.json`: Stores development flags in one place for us to easily change when testing
- `plot_trajectory.py`: Previous matplotlib code to plot the simulated trajectory. Added specifications to all functions and made global variables to share mission and other variables across animation functions because FuncAnimation doesn't let you send in arguments for init (or at least i couldn't find a way to do so)
- `parser.py`: Implementation to read command-line arguments (or optionally default to the values we had in sim_trajectory) that is used for our user-defined arguments needed to initialize our robot and mission. This will be used to run our script for the robot from the GUI. Note: only lawnmower control mode is accounted for since we need to see if guided_waypoints works still.
- `transmission.py`: Previous packet thread code. Added specification, updated loop condition, and removed interfile csv communication (since our gui is now on a separate codebase). Real implementation will depend on our transmission communication (e.g. RF module vs LARA)
- `sim_trajectory.py` --> `main.py` **[RENAME]**: Load in config flags and command-line arguments to initialize our mission and robot. Import and incorporate functions from all of the new files listed above ^. Nest code for packet transmission and plotting under respective config flags. Updated logic to set `is_sim`, `init_control_mode` (temporary - this should be completely user defined in the future),  and `phase` (in `robot_state.py`).
- Updated tests accordingly

## Test Plan

Ran `python -m engine.main` w/ diff configs settings
Ran `python -m tests.compilation_tests.test_runner`
Ran `python -m engine.main 10 10` w/ simulate_trajectory flag true in config.json (robot should be in a different starting location
Screen Shot 2023-04-20 at 1 31 06 AM

## Other

Future TODOs:
- Add additional control modes to `parser.py` once we figure out which control modes are currently functional or not. (e.g. guided_waypoints working w move_to_target nodes, do we need all of lawnmower and lawnmower_a and lawnmower_b)
- Replace database packet making implementation with a new packet making implementation using robot state and mission state


engine/main.py
```diff
@@ -0,0 +1,47 @@
...
+if __name__ == "__main__":
+    # Load config from JSON file
+    with open(ENGINE_PATH +"/config.json", "r") as fp:
+        config_args = json.load(fp)
+    
+    # Load user-defined arguments
+    user_args = parse_main()
+
+    # Development config flags
+    is_transmit = config_args.get("is_transmit") # Set to true when the rpi/robot is communicating w/ the GUI
```
Aayush-Agnihotri, line 26: remove is_transmit and just use config_args.get("is_transmit") wherever is_transmit is

engine/parser.py
```diff
@@ -0,0 +1,38 @@
...
+   parser.add_argument("base_lat", type=float, nargs='?', const=1, default=42.444250,
+                    help="a float for the base stations's latitude coordinate")
+   parser.add_argument("base_long", type=float, nargs='?', const=1, default=-76.483682,
+                    help="a float for the base stations's longitude coordinate")
+   parser.add_argument("-l", "--lawnmower", action="store_false",
+                    help="Set intial control mode to lawnmower") # TODO: add for additional control modes, after centralizing the different lawnmowers
+   args = parser.parse_args()
+   return(vars(args))
+
+parse_main()
```
Aayush-Agnihotri, line 38: Is this line needed? Idrk anything about command line parsing

engine/robot_state.py
```diff
@@ -93,8 +93,8 @@ def __init__(self, **kwargs):
         # If false, only uses GPS and IMU; else, uses EKF
         self.using_ekf = False
         self.is_sim = kwargs.get("is_sim", not is_raspberrypi())
-        self.should_store_data = kwargs.get("should_store_data", False)
-        self.phase = Phase(kwargs.get("phase", Phase.SETUP))
+        self.store_data = kwargs.get("store_data", False)
+        self.phase = Phase(kwargs.get("phase", Phase.TRAVERSE if self.is_sim else Phase.SETUP))
```
Aayush-Agnihotri, line 97: Is there any setup needed for a simulation or no

tests/compilation_tests/database_test.py
```diff
@@ -169,7 +169,7 @@ def test_update_data(self):
                     self.assertEqual(ans, database.get_data(name), name)
 
     def test_phase_as_value(self):
-        robot_state_phase = Robot_State(xpos= 0, ypos =0, heading =0, epsilon =0, max_velocity =0, radius =0)
+        robot_state_phase = Robot_State(xpos= 0, ypos =0, heading =0, epsilon =0, max_velocity =0, radius =0, phase = Phase.SETUP)
```
AC408, line 172: are the params needed if they're the default value?

tests/compilation_tests/database_test.py
```diff
@@ -13,7 +13,7 @@
 class TestDataBase(unittest.TestCase):
     # DataBase instances to test on
     robot_state_default = Robot_State(xpos=0, ypos=0, heading=0, epsilon=0, max_velocity=0,
-                          radius=0)
+                          radius=0, phase = Phase.SETUP)
```
AC408, line 16: are the params needed if they're the default value?

engine/main.py
```diff
@@ -0,0 +1,47 @@
...
+    is_sim = config_args.get("is_sim") if is_raspberrypi() else not is_raspberrypi() # Set to true when simulating the rpi, set to false when running on rpi
+    store_data = config_args.get("store_data") # Set to true when we want to track/store csv data
```
AC408, line 28: if im not mistaken, this is only used to pass into Robot_State. I think it might be more concise to directly have this there?

engine/parser.py
sazhang02: if u mean line 38 it is not i forgot and left it there from testing o7 good catch

engine/plot_trajectory.py
```diff
@@ -0,0 +1,138 @@
...
+        range = m.mission_state.roomba_radius + margin
+        init_x = m.mission_state.base_station.position[0]
+        init_y = m.mission_state.base_station.position[1]
+        plt.xlim([init_x-range, init_x+range])
+        plt.ylim([init_y-range, init_y+range])
+        circle = plt.Circle((init_x, init_y), m.mission_state.roomba_radius)
+        ax.add_patch(circle)
+
+    elif m.mission_state.control_mode != ControlMode.MANUAL:
+        goals = waypoints_to_array(m.mission_state.all_waypoints)
```
AC408, line 117: I don't think this is used anywhere?

AC408 (changes requested):
some comments

engine/robot_state.py
sazhang02: Thats a good question 🤔  i think there are variables like truthpose in robot_state and some waypoint attributes in mission that are only used in the case of a simulation. 

If we want to make it cleaner we could nest it those attributes under the is_sim flag?? I guess this depends if we only want the matplotlib graph to appear/populate that info only in simulated cases tho. Hmmm not sure but we might need this info for the gui to help us plot a live graph?? 🤔 

do u think its worth?

tests/compilation_tests/database_test.py
sazhang02: good point, only the ones that differ from default are needed! i think the ones that are default here are xpos, ypos , heading so ill remove those

tests/compilation_tests/database_test.py
sazhang02: same as above will remove the ones w same value as default o7

engine/main.py
sazhang02: pushed a change for this. i was thinking about passing config arg into robot state but after testing i realized we would need to load the config args in every test which kinda sucks so i passed in the values of is_sim and store_data there not sure if thats the best approach tho?

engine/plot_trajectory.py
sazhang02: hmm yeah its greyed out on mine too i guess it wasnt used before as well? will remove

engine/main.py
sazhang02: https://github.com/cornellnexus/src/pull/131#discussion_r1174086868

jinlin270 (commented):
The is_sim flag is true, but when I ran main there is no graph.

AC408:
> The is_sim flag is true, but when I ran main there is no graph. Does anyone else have this issue?

what did you run?

jinlin270:
> > The is_sim flag is true, but when I ran main there is no graph. Does anyone else have this issue?
> 
> what did you run?

I ran python -m engine.main and python -m engine.main 10 10, neither produced the animation

AC408:
> > > The is_sim flag is true, but when I ran main there is no graph. Does anyone else have this issue?
> > 
> > 
> > what did you run?
> 
> I ran python -m engine.main and python -m engine.main 10 10, neither produced the animation

oh yeah, same

engine/parser.py
```diff
@@ -0,0 +1,36 @@
...
+                    help="a float for the robot's initial x position in coordinate space")
+   parser.add_argument("ypos", type=float, nargs='?', const=1, default=0,
+                    help="a float for the robot's initial y position in coordinate space")
+   parser.add_argument("heading", type=float, nargs='?', const=1, default=math.pi/4,
+                    help="a float for the robot's initial heading")
+   parser.add_argument("base_lat", type=float, nargs='?', const=1, default=42.444250,
+                    help="a float for the base stations's latitude coordinate")
+   parser.add_argument("base_long", type=float, nargs='?', const=1, default=-76.483682,
+                    help="a float for the base stations's longitude coordinate")
+   parser.add_argument("-l", "--lawnmower", action="store_false",
```
AC408, line 33: this doesn't work :(

AC408: the -l flag

AC408 (changes requested):

jinlin270 (approved):
Great PR!

engine/robot_state.py
Aayush-Agnihotri: in that case, its probably fine how it is cuz we might want to also have a graph while testing on the robot idk

Aayush-Agnihotri (approved):

engine/robot_state.py
sazhang02: omg we just talked about having a graph on the robot yesterday thats wild

engine/parser.py
sazhang02: oops yeah so i made it such that if u do -l it turns it off instead of on LOL gonna just leave this as a todo bc we dont know what control modes we have available x.x

AC408 (changes requested):
If I try to run `python -m engine.main 10 10 2 100 200`, nothing happens

sazhang02:
oh no :0

sazhang02:
i turned the transmit (aka packet thread) off so i think it seems like nothing is happening bc theres nothing being printed anymore when mission is running and since the values are like weird for our mission to run (e.g. huge lat long bounds for grid) it executes for a long time it just runs forever (if u turn on transmit in config it prints a bunch of stuff).

Running `python -m engine.main 10 10 0.79 42.444250 -76.483682` (using our current grid of eng quad) works for me and i think w other lat, longs not working thats more of an issue w mission?

@AC408

AC408 (approved):

PR #135, opened 2023-09-09
Fixing PID traversal bug

AC408:
## References
- Documentation: https://docs.google.com/document/d/1w-NiddRpPD56dYxc8AsbNXfKp-4ZWtESaPwqZGfKcz4/edit?usp=sharing

## Summary
- In Kinematics, in integrate_odom(), removed domain constraint of new_theta (which was (-2*pi, 2*pi)
- In traverse_standard_test, added the test cases that gave us the errors
- Changed move_to_target_node to turn towards the desired goal, then have the error be the euclidean distance for moving PID (not turning)
- removed rounding of state

### Description
- When setting angle for heading PID to 0, the robot, when moving, could go from a heading of 0 to 2pi - epsilon for a small epsilon, which causes the robot to turn a full circle to get back to a heading of 0
- Added flag to use heading_pid vs original code; set using_heading_pid to true to use the new logic with turning first before moving
- Added flag to visualize move_to_target_node; set simulate to true to show the visualization

## Test Plan
- python -m tests.compilation_tests.test_runner and checking the visualization

### DIFF
--- a/engine/kinematics.py
+++ b/engine/kinematics.py
@@ -21,8 +21,8 @@ def robot_to_global(pose, x_robot, y_robot):
     px = pose[0]
     py = pose[1]
     theta = pose[2]
-    Tgr = np.array([[math.cos(theta), -1 * math.sin(theta), px], \
-                    [math.sin(theta), math.cos(theta), py], \
+    Tgr = np.array([[math.cos(theta), -1 * math.sin(theta), px],
+                    [math.sin(theta), math.cos(theta), py],
                     [0, 0, 1]])
     # DEBUG UPDATE: added [1] to array in order to make it 3x3 times 3x1 rather than 2x1
     # rounding x,y coordinates to make it testable (ex: 4.000000001 --> 4); noise?
@@ -34,7 +34,7 @@ def robot_to_global(pose, x_robot, y_robot):
 def global_to_robot(pose, x_global, y_global):
     """
     Transforms xy_global into robot coordinates. Outputs a [2x1] np array. 
-    
+
     Inputs: 
         pose: robot's current pose (global) [3x1]
         x_global: x coordinate in global frame 
@@ -45,8 +45,8 @@ def global_to_robot(pose, x_global, y_global):
     px = pose[0]
     py = pose[1]
     theta = pose[2]
-    Tgr = np.array([[math.cos(theta), -1 * math.sin(theta), px], \
-                    [math.sin(theta), math.cos(theta), py], \
+    Tgr = np.array([[math.cos(theta), -1 * math.sin(theta), px],
+                    [math.sin(theta), math.cos(theta), py],
                     [0, 0, 1]])
     Trg = np.linalg.inv(Tgr)
     # DEBUG UPDATE: Adjusting parameters to be consistent with first funtion
@@ -66,7 +66,8 @@ def feedback_lin(curr_pose, vx_global, vy_global, epsilon):
     """
     theta = float(curr_pose[2])
     v = vx_global * math.cos(theta) + vy_global * math.sin(theta)
-    w = (-1 / epsilon) * vx_global * math.sin(theta) + (1 / epsilon) * vy_global * math.cos(theta)
+    w = (-1 / epsilon) * vx_global * math.sin(theta) + \
+        (1 / epsilon) * vy_global * math.cos(theta)
     return (v, w)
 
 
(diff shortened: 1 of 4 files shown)


engine/robot_logic/traversal.py
```diff
@@ -64,17 +68,34 @@ def move_to_target_node(robot_state, target, allowed_dist_error, database):
     robot_state.goal_location = target
     robot_state.loc_pid_x.reset_integral()
     robot_state.loc_pid_y.reset_integral()
+    using_heading_pid = True
+    simulate = False
+    if simulate:
+        i = 0
+    import matplotlib.pyplot as plt
```
Shreya1233, line 75: why is the import statement here?

engine/robot_logic/traversal.py
```diff
@@ -64,17 +68,34 @@ def move_to_target_node(robot_state, target, allowed_dist_error, database):
...
     while (distance_away > allowed_dist_error) and not (robot_state.phase == Phase.AVOID_OBSTACLE):
         # Error in terms of latitude and longitude, NOT meters
         x_coords_error = target[0] - robot_state.state[0]
         y_coords_error = target[1] - robot_state.state[1]
+        if simulate:
+            i += 1
+            if i == 50:
```
Shreya1233, line 82: why is i = 50 resulting in the break?

engine/robot_logic/traversal.py
```diff
@@ -64,17 +68,34 @@ def move_to_target_node(robot_state, target, allowed_dist_error, database):
...
+                break
+        desired_angle = math.atan2(
+            target[1] - robot_state.state[1], target[0] - robot_state.state[0])
 
         x_vel = robot_state.loc_pid_x.update(x_coords_error)
         y_vel = robot_state.loc_pid_y.update(y_coords_error)
+        if using_heading_pid:
+            turn_to_target_heading(robot_state, desired_angle, 0.01, database)
```
Shreya1233, line 90: unrelated to PIR - currently I know we are doing point turn. However, whenever we decide on turning mechanism, turn_to_target_heading will probably have to be revamped.

tests/compilation_tests/robot_test/robot_test.py
```diff
@@ -44,8 +44,10 @@ def test_travel(self):
         new_x = round(-1.3662, 3)
         new_y = round(-0.366198, 3)
         new_theta = round(3 * math.pi / 2, 3)
+        rounded_robot_one_state = [round(state[0], 3)
```
Shreya1233, line 47: why do we round it again after we rounded x, y, and theta?

jinlin270 (commented):
Great PR Allen :D

tests/compilation_tests/robot_test/traverse_standard_test.py
```diff
@@ -15,11 +15,46 @@
 gps_noise_range = .3
 
 
+class TestMoveToTargetNode(unittest.TestCase):
```
Shreya1233, line 18: wanted to check if these are the existing test cases that we were failing originally?

Shreya1233 (changes requested):
Hey Allen! For the most part everything looks fine. Just had a couple of clarification questions/comments + wanted to know why we removed the constraint for integrate_odom.

engine/robot_logic/traversal.py
AC408: it's not needed elsewhere. If we're only using it in move_to_target_node, I dunno if we need to import it globally

engine/robot_logic/traversal.py
AC408: we're limiting the number of points we are plotting. If we don't limit it, then the graphing would take forever because there is a lot of iterations

engine/robot_logic/traversal.py
AC408: yep

tests/compilation_tests/robot_test/robot_test.py
AC408: to make the test case work; the previous test case has new_x, new_y, and new_theta be rounded because our travel implementation rounds the state.

engine/kinematics.py
```diff
@@ -114,9 +115,11 @@ def integrate_odom(pose, d, phi):
...
         new_theta = pose[2]
     else:
-        new_x = pose[0] + (d / phi) * (math.sin(pose[2] + phi) - math.sin(pose[2]))
-        new_y = pose[1] + (d / phi) * (-math.cos(pose[2] + phi) + math.cos(pose[2]))
-        new_theta = (pose[2] + phi) % (2 * math.pi)
+        new_x = pose[0] + (d / phi) * \
+            (math.sin(pose[2] + phi) - math.sin(pose[2]))
+        new_y = pose[1] + (d / phi) * \
+            (-math.cos(pose[2] + phi) + math.cos(pose[2]))
+        new_theta = pose[2] + phi
```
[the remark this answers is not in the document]
AC408: if we imagine the robot desired angle set to 0, then if the robot is at theta = 0 and turn left slightly, the angle will be 0 - epsilon, which would have (with the constraint) turned into 2 * pi - epsilon, which is not what we want. This is the same with theta = 2 *pi instead of 0

tests/compilation_tests/robot_test/traverse_standard_test.py
AC408: I don't think we failed any, just had unexpected behavior

jinlin270 (approved):
Everything looks good!

Shreya1233 (approved):
good pr :)

PR #140, opened 2023-09-20
Fix traversal algorithm bug

AC408:
## Summary
- Traverse_standard previously had a while loop, so even if the robot detects an obstacle and set the avoid obstacle flag, the robot would continue doing traverse_standard until there is no more waypoints left

### Description
- Changed logic for traverse_standard: since we effectively already have a while loop in mission, we can remove the while loop from traverse_standard and have the robot go through all waypoints via mission’s while loop instead
- Updated test cases to reflect new logic
- Added logic for doing obstacle avoidance
- Removed unnecessary comments

## Test Plan
- called `python3 -m tests.compilation_tests.test_runner`

### DIFF
--- a/engine/robot_logic/traversal.py
+++ b/engine/robot_logic/traversal.py
@@ -8,23 +8,25 @@ from engine.kinematics import *
 from engine.robot_logic.robot_helpers import phase_change, calculate_dist
 from csv_files.csv_util import write_state_to_csv
 
+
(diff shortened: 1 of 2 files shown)


jinlin270 (approved):
Great PR

Shreya1233 (approved):
Changes look good to me! Good PR.

PR #143, opened 2023-10-18
Implementing break beam sensors 

Andrew-Lowitt:
## Summary
- Initialized the breakbeam sensors in sensor_modules.py
- Added the initialized sensors to robot_state.py
- For the function traverse_standard in traversal.py I added a guard that makes sure the bucket is not full. If the bucket is full, then the phase will be set to return

### Description
Followed the implementation style of other sensors in the sensors_module and robot_state file
## Test Plan
Break_Beam has been tested physically with the actual sensors, so we can ensure that this is functioning properly

## Other
I am not sure that the method of implementing the sensors into the sensor_module and robot_state.py is actually correct. Again, I just followed the style of how other sensors were implemented

### DIFF
--- a/engine/robot_logic/traversal.py
+++ b/engine/robot_logic/traversal.py
@@ -7,6 +7,7 @@ from constants.definitions import *
 from engine.kinematics import *
 from engine.robot_logic.robot_helpers import phase_change, calculate_dist
 from csv_files.csv_util import write_state_to_csv
+from electrical import breakbeam
 
 def traversal_logic(robot_state, mission_state, database):
         """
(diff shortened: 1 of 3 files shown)


AC408 (changes requested):
Can you fix the merge conflict and the broken test(s)?

Shreya1233 (changes requested):
When I run the test runner (python -m tests.compilation_tests.test_runner) and the main script (python -m tests.compilation_tests.test_runner), I'm getting errors. Could you fix this and then re-request review?

PR #146, opened 2023-10-24
Fix roomba mode

AC408:
## Summary
Changed roomba mode to use PID to move instead of time-based motions. Updated simulation to work for roomba mode.

### Description
- At every instance we have a loop and dont have a sleeping time when not simulating, add one. 
- At every instance where we use spin_motors, made sure there is a corresponding spin_motors(0, 0), so motors will stop spinning. 
- Changed logic so prediction of whether robot will exit certain range is based on actual robot velocity. Added refactoring to make some traverse_roomba motion to be PID-based. 
- In grid.py, changed LAWNMOWER_GUIDED to LAWN_MOWER_A. 
- In parser, added argument for initializing control mode. 
- In main, added parser argument for control mode and initialized mission state and robot state with the control mode. 
- In traversal, updated traverse_roomba return. In traversal, updated traverse_standard base case. 
- In traversal, limited target_heading in turn_to_target-heading to reflect new range for theta.

## Test Plan
- `python -m tests.compilation_tests.test_runner`
- `python -m engine.main` 
- `python -m engine.main -c lawnmower`

## Other
TODO: add the other control modes to parser


engine/robot_logic/traversal.py
```diff
@@ -208,64 +212,98 @@ def update_state(robot_state, velocity, omega):
...
+    dt = 10
+    accumulated_time = 0
     # TODO: battery_limit, time_limit, tank_capacity is full
     exit_boolean = False
-    from electrical.ultrasonic_sensor import Ultrasonic
+    if not robot_state.is_sim:
+        from electrical.ultrasonic_sensor import Ultrasonic
 
-    front_ultrasonic = Ultrasonic(0)
+        front_ultrasonic = Ultrasonic(0)
```
Shreya1233, line 234: don't we already do this in robot setup_logic? Why can't we just directly use the front_ultrasonic in robot_state?

engine/robot_logic/traversal.py
```diff
@@ -208,64 +212,98 @@ def update_state(robot_state, velocity, omega):
...
         roomba_radius (Double): Maximum radius from the base station that the robot can move
     Returns:
-        None
+        robot_state
     """
     # to add obstacle avoidance, can either change roomba to go to point on the circle in the direction of the
     # robot direction or use raw sensor value
     # can we guarantee that when giving a velocity that the robot will move at that velocity?
-    dt = 0
+    dt = 10
```
Shreya1233, line 227: why is dt = 10?

AC408: could be any value. its just what we're passing to time.sleep(). smaller value means update faster but also hog cpu more. i think 10 is usually standard. at least we've always been using time.sleep(10) in our while loops

engine/robot_logic/traversal.py
```diff
@@ -208,64 +212,98 @@ def update_state(robot_state, velocity, omega):
...
             return None
-        curr_x = robot_state.state[0]
-        curr_y = robot_state.state[1]
-        curr_head = robot_state.state[2]
-        [new_x, new_y, new_theta] = robot_state.ekf.get_predicted_state(
-            [curr_x, curr_y, curr_head],
-            [robot_state.move_dist * robot_state.time_step, 0],
-        )
+        curr_pos = robot_state.state
+        vel = calculate_dist(past_pos[:2], curr_pos[:2]) / (dt / 1000)
```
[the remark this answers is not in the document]
AC408: dt is in ms, we want val to be m/s, not m/ms

engine/robot_logic/traversal.py
```diff
@@ -208,64 +212,98 @@ def update_state(robot_state, velocity, omega):
...
-        is_next_timestep_blocked = next_radius < robot_state.detect_obstacle_range
+        if not robot_state.is_sim:
+            # if obstacle detected; have to make sure move_dist < detect_obstacle_range; otherwise, robot might hit obstacle before we go back to this loop
+            is_next_timestep_blocked = (
+                front_ultrasonic.distance() < robot_state.detect_obstacle_range
+            )
+        else:
+            next_radius_2 = calculate_dist(base_station_loc, curr_pos[:2])
+            threshold = 1
+            is_next_timestep_blocked = abs(next_radius_2 - roomba_radius) < threshold
```
Shreya1233, line 262: the changes to is_next_timestep_blocked make sense but why don't we also want to check for the old case: next_radius < robot_state.detect_obstacle_range? don't we want to check if our future movement will result in moving through the obstacle? I know we check for next_radius > roomba_radius in the line below but I think this is a separate case right?

engine/robot_logic/traversal.py
```diff
@@ -208,64 +212,98 @@ def update_state(robot_state, velocity, omega):
...
+                robot_state,
+                curr_pos[2] + robot_state.turn_angle,
+                allowed_heading_error,
+                database,
+            )
         else:
-            if robot_state.is_sim:
-                travel(robot_state, robot_state.move_dist, 0)
-            else:
+            travel(robot_state, robot_state.move_dist, 0)
```
Shreya1233, line 285: in this else clause aren't we supposed to move toward the goal location (robot_state.goal_location = (new_x, new_y))? not sure if we should use move_to_target_node instead or if the changes you're making to travel will take this into account?

Shreya1233 (changes requested):
Looks good for the most part! Just had a few questions.

engine/robot_logic/traversal.py
Shreya1233: oh yeah I just saw later on dt was also used in time.sleep(). That makes sense!

engine/robot_logic/traversal.py
AC408: we aren't moving to the goal location, we just want to move forward. Using move_to_target_node doesn't really make sense because we're just moving forward; there is no end goal. We could set a target to be a few "insert length" in front of us, but using PID on every single instance of that length is slow (robot accelerates, then decelerates, etc)

Shreya1233: so with Roomba mode (and also with the travel method) we are disregarding the idea of moving from waypoint to waypoint? Then we should prolly remove this line: robot_state.goal_location = (new_x, new_y).  Also why do we use new_x and new_y to check whether or not the robot would crash into an obstacle/is within the Roomba radius if the robot isn't even moving to this new position?

AC408: yeah, i could remove goal_location, but do we want to not update it? if we dont, it no longer "keep its invariant" or whatever. new_x and new_y is where I predict the robot to be, not necessarily where it will go, and also not where it should be (EKF). Otherwise, if we cant predict where the robot will be, we cant determine if it will go out of bounds until its already out of bounds

Shreya1233 (approved):
looks good to me!

jinlin270 (approved):

jinlin270:
Great PR! Everything looks good

PR #148, opened 2023-11-01
Created a Simplified move_to_target_node method

polarcow285:
### Description
- Copied `move_to_target_node` and removed any references to PID, EKF, and simulation.
- Instead of using a `predicted_state` from an EKF, the code uses the robot's current state instead.
- Used `x_coords_error` and `y_coords_error` directly `x_vel` (x velocity), `y_vel` (y velocity). Before, the velocities came from PID computations.

## Test Plan

- Run `python -m tests.compilation_tests.robot_test.traverse_standard_test`
- The tests perform the same functionality as `move_to_target_node` tests.


csv_files/phases.csv
```diff
@@ -0,0 +1,65 @@
+Phase.TRAVERSE
```
AC408, line 1: remove changes in this file; better yet, add this file to the gitignore file

engine/robot_logic/traversal.py
```diff
@@ -50,6 +50,54 @@ def traverse_standard(robot_state, unvisited_waypoints, allowed_dist_error, data
...
+    distance_away = calculate_dist(target, robot_state.state)
+    robot_state.goal_location = target
+    
+    while (distance_away > allowed_dist_error) and not (robot_state.phase == Phase.AVOID_OBSTACLE):
+        # Error in terms of latitude and longitude, NOT meters
+        x_coords_error = target[0] - robot_state.state[0]
+        y_coords_error = target[1] - robot_state.state[1]
+
+        desired_angle = math.atan2(
+            target[1] - robot_state.state[1], target[0] - robot_state.state[0])
```
AC408, line 75: nit, duplicate code

engine/robot_logic/traversal.py
```diff
@@ -50,6 +50,54 @@ def traverse_standard(robot_state, unvisited_waypoints, allowed_dist_error, data
...
+            robot_state.state, x_vel, y_vel, robot_state.epsilon)
+
+        # clamping of velocities:
+        (limited_cmd_v, limited_cmd_w) = limit_cmds(
+            cmd_v, cmd_w, robot_state.max_velocity, robot_state.radius)
+
+        robot_state.linear_v = limited_cmd_v[0]
+        robot_state.angular_v = limited_cmd_w[0]
+        travel(robot_state, robot_state.time_step *
+               limited_cmd_v[0], robot_state.time_step * limited_cmd_w[0])
```
AC408, line 90: is this based on the new move_to_target_node logic or the old one?

engine/robot_logic/traversal.py
```diff
@@ -50,6 +50,54 @@ def traverse_standard(robot_state, unvisited_waypoints, allowed_dist_error, data
...
+        cmd_v, cmd_w = feedback_lin(
+            robot_state.state, x_vel, y_vel, robot_state.epsilon)
+
+        # clamping of velocities:
+        (limited_cmd_v, limited_cmd_w) = limit_cmds(
+            cmd_v, cmd_w, robot_state.max_velocity, robot_state.radius)
+
+        robot_state.linear_v = limited_cmd_v[0]
+        robot_state.angular_v = limited_cmd_w[0]
+        travel(robot_state, robot_state.time_step *
```
AC408, line 89: TODO: we need to unify time.sleep and robot_state.time_step

engine/robot_logic/traversal.py
```diff
@@ -50,6 +50,54 @@ def traverse_standard(robot_state, unvisited_waypoints, allowed_dist_error, data
...
+               limited_cmd_v[0], robot_state.time_step * limited_cmd_w[0])
+        if not robot_state.is_sim:
+            robot_state.motor_controller.spin_motors(
+                limited_cmd_w[0], limited_cmd_v[0])
+            time.sleep(10)
+            
+        # location error (in meters)
+        distance_away = calculate_dist(target, robot_state.state)
+        
+
```
AC408, line 99: do we want to update database?

tests/compilation_tests/robot_test/traverse_standard_test.py
```diff
@@ -26,7 +26,19 @@ def test1(self):
         allowed_dist_error = 0.05
         move_to_target_node(r2d2_state,
                             (42.444250, -66.483682), allowed_dist_error, database)
+    
+    #Runs test1 using simple_move_to_target_node
+    def test1_simple(self):
+        r2d2_state = Robot_State(xpos=42.444250, ypos=-76.483682, heading=0,
+                                 epsilon=0.2, max_velocity=0.5, radius=0.2, phase=2, position_kp=.1, heading_kp=.1, position_kd=0.0, heading_kd=0.0
```
AC408, line 33: iirc, most of these params have default args. Do we need to specify some of these args for testing purposes?

AC408: i also see this defined in test1. i think it would be better to initialize these r2d2 within the class itself to remove duplicate code

tests/compilation_tests/robot_test/traverse_standard_test.py
```diff
@@ -47,6 +70,16 @@ def test3(self):
...
+                                 )
+        r2d2 = Robot(robot_state=r2d2_state)
+        database = DataBase(r2d2)
+        allowed_dist_error = 0.5
+        simple_move_to_target_node(r2d2_state,
+                            (43.444250, -76.483682), allowed_dist_error, database)
 
 # Takes approx. 9 seconds for 5 tests
```
AC408, line 84: lets remove this as well

AC408 (changes requested):
Looks good. Similar to Ahan's task, I'm assuming we are only using simpler_move_to_target_node for testing our velocity calculation and error? If so, I think it would make more sense to add this to a testing file different from robot_logic

engine/robot_logic/traversal.py
polarcow285: This code was based off of the original move_to_target_node

tests/compilation_tests/robot_test/traverse_standard_test.py
polarcow285: What exactly should I remove? The entire test3? (did you highlight a specific part of the code? if you did, i don't think i see it)

engine/robot_logic/traversal.py
polarcow285: Added a todo for this in the code.

engine/robot_logic/traversal.py
polarcow285: @Shreya1233 Should we update the database here?

tests/compilation_tests/robot_test/traverse_standard_test.py
polarcow285: Made class variables and removed duplicate code.

engine/robot_logic/traversal.py
AC408: uhhh, i dont know if we want to use the old logic, but ig we cant use the new logic too, since it relies on turn_to_target_node; I'm actually not sure but it might be a good idea to ask Shreya about this

tests/compilation_tests/robot_test/traverse_standard_test.py
AC408: duplicate code, unnecessary args

polarcow285: Duplicate code and unnecessary args removed via setUp method

engine/robot_logic/traversal.py
polarcow285: Todo: keep as original logic. when we test it out, and doesnt meet the intedned tracjeoty, then add back alan's pid fix.

engine/robot_logic/traversal.py
polarcow285: Added a todo for this question.

engine/robot_logic/traversal.py
polarcow285: Added this todo.

PR #149, opened 2023-11-04
Added simpler turn to target method and tests

ahanbm:
## Summary
Added a simpler version of the turn to target heading function that doesn't rely on the PID. Copied test cases from sensor_functionality_tests and modified them to test the simpler turn function.

### Description

The original turn function was left unchanged and a new simpler turn function was added to traversal.py. The original function was too complicated because it required a PID and used functions that were unnecessary for simple traversals. 
Old Method: https://drive.google.com/file/d/1N3UA-pxY_fdw9TrUNMhzvM-ib3dpfZ_H/view?usp=sharing
New Method: https://drive.google.com/file/d/1Q-plhakmphXIcNF7dal6mg9c9SblEFiu/view?usp=sharing

I shortened the method and removed the PID. The main decision I had to make is what to assign the intended angular velocity. I ended up equating it to the theta_error so that it was a simple linear approximation. 

## Test Plan

I copied the sensor_physical_testing file and adapted it so that it tested my simpler turn method. I modified the new file until it had no files. Because the tests are functionality tests I can't tell if the method works till it is tested physically but it does run with no errors when simulated. I added 5 separate test cases for the turn method, turning by -180, -90, 0, 90, and 180 degrees. I had to change a few lines of sensor_physical_test to get it to work, the most important being the initialization of the robot state which I inputted as a dictionary rather than 6 separate inputs.


Shreya1233 (commented):
Gen comment: there was a bug in our code which got merged in main. I think since you pulled main with the bug, we can't run the main file unfortunately. Allen's new PR fixes the problem, so once we merge that PR into main, you should merge main into your branch so the bug's no longer there.

engine/robot_logic/traversal.py
```diff
@@ -308,3 +308,36 @@ def turn_to_target_heading(
...
+    Turns robot in-place to target heading + or - allowed_heading_error
+        target_heading: the heading in radians the robot should approach at the end of in-place rotation.
+        allowed_heading_error: the maximum error in radians a robot can have to target heading while turning in
+            place.
+    """
+    # we dont want the robot to avoid obstacle here
+    robot_state.enable_obstacle_avoidance = False
+
+    while abs(target_heading - float(robot_state.state[2])) > allowed_heading_error:
+        theta_error = target_heading - robot_state.state[2]
```
AC408, line 326: TODO: we can probably optimize this such that the robot turns the shorter angle (imagine target_heading is 359 degrees and robot_state.state[2] is 1, then it is more optimal to turn from 359 to 360 to 1 than 359 to 358 ... to 1

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,32 @@
+# THIS IS A TESTING SCRIPT FOR TURN_TO_TARGET_HEADING. USE THIS SCRIPT WHEN PHYSICALLY TESTING THE ROBOT AND CHANGE THE test_state TO THE STATE YOU NEED
+
+from engine.robot_state import Robot_State
+from engine.robot import Robot
+from engine.robot_logic.traversal import simpler_turn_to_target_heading
+from engine.database import DataBase
+import numpy as np
```
AC408, line 7: is this import necessary?

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,32 @@
...
+        simpler_turn_to_target_heading(rb_state, math.pi / 2, 2, database)
+
+    elif test_state == 2:
+        simpler_turn_to_target_heading(rb_state, math.pi, 2, database)
+    
+    elif test_state == 3:
+        simpler_turn_to_target_heading(rb_state, -math.pi / 2, 2, database)
+    
+    elif test_state == 4:
+        simpler_turn_to_target_heading(rb_state, -math.pi, 2, database)
```
AC408, line 32: to be more comprehensive, we could also change starting position x,y

AC408 (changes requested):
Looks good. I'm assuming we are only using simpler_turn_to_target_heading for testing our velocity calculation and error? If so, I think it would make more sense to add this to testing file different from robot_logic

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,32 @@
...
+from engine.robot import Robot
+from engine.robot_logic.traversal import simpler_turn_to_target_heading
+from engine.database import DataBase
+import numpy as np
+import math
+
+if __name__ == "__main__":
+    # 0 does nothing, 1 is a 90 degree turn, 2 is a 180 degree turn, 3 is a -90 degree turn, 4 is a negative 180 degree turn
+    test_state = 0
+    rb_state = Robot_State(x_pos=0, y_pos = 0, heading = 0, epsilon = 0.2, max_velocity = 0.5, radius = 0.2)
```
AC408, line 13: i dont think we need to specify some of these params

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,32 @@
...
+from engine.robot_state import Robot_State
+from engine.robot import Robot
+from engine.robot_logic.traversal import simpler_turn_to_target_heading
+from engine.database import DataBase
+import numpy as np
+import math
+
+if __name__ == "__main__":
+    # 0 does nothing, 1 is a 90 degree turn, 2 is a 180 degree turn, 3 is a -90 degree turn, 4 is a negative 180 degree turn
+    test_state = 0
```
AC408, line 12: you should make it such that all of the test are ran when the file is ran, so test_state should iterate. also, nit, this should be made into an actual unittest file

ahanbm: The reason my file is formatted like this is because sensor_physical_testing was the only file in the codebase testing the turn function so I copied it. I will reformat my tests because I think it's a good suggestion but it might take some time.

tests/functionality_tests/simple_turn_test.py
AC408: i dont think this has been addressed yet

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,134 @@
...
+        database.update_data(
+            "state", robot_state.state[0], robot_state.state[1], robot_state.state[2]
+        )
+
+    # re-enable after finishing turning
+    robot_state.enable_obstacle_avoidance = True
+
+
+target = [0, 0]
+add_to_x = False
```
AC408, line 48: unused var, remove

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,134 @@
...
+gps_noise_range = 0.3
```
AC408, line 49: unused var, remove

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,134 @@
...
+        
+        database.update_data(
+            "state", robot_state.state[0], robot_state.state[1], robot_state.state[2]
+        )
+
+    # re-enable after finishing turning
+    robot_state.enable_obstacle_avoidance = True
+
+
+target = [0, 0]
```
AC408, line 47: i dont think this is used, remove

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,134 @@
...
+add_to_x = False
+gps_noise_range = 0.3
+
+
+class TestMoveToTargetNode(unittest.TestCase):
```
AC408, line 52: for each test case, you should be asserting that the final heading you got from calling simpler_turn_to_target_heading is what you expected

tests/functionality_tests/simple_turn_test.py
```diff
@@ -0,0 +1,134 @@
...
+        database = DataBase(r2d2)
+        allowed_dist_error = 0.1
+        simpler_turn_to_target_heading(
+            r2d2_state, math.pi / 2, allowed_dist_error, database
+        )
+
+# Takes approx. 9 seconds for 5 tests
+
+if __name__ == "__main__":
+    unittest.main()
```
AC408, line 134: add TODO: make a physical script and document expected behavior. Set a bool to toggle using the physical script vs the unit test

AC408 (changes requested):
Everything looks good, but there is just a few small issues that needs to be touched up.

AC408 (approved):
Looks good!

span1012 (approved):

PR #150, opened 2023-11-04
Traverse directions

jinlin270:
## Summary
This PR enables traversal in UP, DOWN, LEFT, and RIGHT directions, rather than just UP and RIGHT directions. 

### Description
Previously the algorithm assumes to traverse in the UP or RIGHT direction, and user is allowed to choose between these 2 using the is_vertical traversal flag. The algorithm is added on slightly so the user can instead choose between 4 directions of traversal, and the find_border_node function is expanded to find the start traversal position to support the different cases.

## Test Plan

This can be tested by running:
for shapes, input one of [ triangle ;  circle ; rectangle ] as shapes (must be spelled correctly and in lower case)
for direction, input one of [UP , DOWN, LEFT, RIGHT]  (must be spelled correctly and in upper case) 

## Other

You can find what the test cases should look like inside 
'src/tests/functionality_tests/activate_shape_graphs'


tests/functionality_tests/activate_shape_graphs/up_triangle.png
AC408, line 1: Is this meant to still be broken?

tests/functionality_tests/activate_shape_graphs/down_triangle.png
AC408, line 1: bottom left looks messed up

engine/grid.py
```diff
@@ -331,36 +337,78 @@ def find_border_nodes(self):
...
+            node = self.nodes[l_row][col]
+            if col > top_left_pos[1]:
+                if node.is_active and self.is_on_border(
+                    (l_row, col), self.num_rows, self.num_cols
+                ):
+                    top_left_pos = (l_row, col)
+                if node.is_active and self.is_on_border(
+                    (r_row, col), self.num_rows, self.num_cols
+                ):
+                    top_right_pos = (r_row, col)
```
AC408, line 402: I dont think this works if the shape is wide on the bottom and narrow at the top (like a triangle)

engine/grid.py
```diff
@@ -331,36 +337,78 @@ def find_border_nodes(self):
...
             self.direction == self.Direction.LEFT
             or self.direction == self.Direction.RIGHT
         ):
             is_vertical = False
+
         border_list = []
         leftmost_node = None
-        leftmost_node_pos = None
-
+        bottom_left_pos = None
```
AC408, line 350: alternative solution is to DFS and keep track of bottom left, bottom right, top left, top right. we should also define "bottom_left", "bottom_right", "top_left", and "top_right" for future purposes if we have an irregular shape (think diamond)

engine/grid.py
```diff
@@ -691,21 +739,35 @@ def get_all_guided_lawnmower_waypoints_adjustable(self, is_vertical):
...
-        if is_vertical:
-            self.direction = self.direction.UP
+        if direction == "DOWN":
+            self.direction = self.Direction.DOWN
+        elif direction == "UP":
+            self.direction = self.Direction.UP
+        elif direction == "LEFT":
+            self.direction = self.Direction.LEFT
+        else:
+            self.direction = self.Direction.RIGHT
```
AC408, line 751: might be better to have `elif direction == "RIGHT"` and then have an else condition that throws an error

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -17,9 +17,16 @@ def main():
...
+    direction = input("Traversal Direction: ")
+    print(direction)
+    if direction == "UP" or direction == "DOWN":
         grid.direction = grid.Direction.UP
+    elif direction == "DOWN":
+        grid.direction = grid.Direction.DOWN
+    elif direction == "RIGHT":
+        grid.direction = grid.Direction.RIGHT
+    else:
+        grid.direction = grid.Direction.LEFT
```
AC408, line 29: same as above, probably better to have elif and else and throw error

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -17,9 +17,16 @@ def main():
         print("flip rows, cols for traversal")
 
     activation_type = input("Activation type: ")
-    is_vertical = "True" == input("Vertical traversal: ")
-    if is_vertical:
+    direction = input("Traversal Direction: ")
+    print(direction)
```
AC408, line 21: add debug state

AC408: and add additional print text for better clarification

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -35,7 +42,7 @@ def main():
             rec_row_start, rec_col_start, rec_row_limit, rec_col_limit
         )
         grid.find_border_nodes()
-        way_points = grid.get_all_guided_lawnmower_waypoints_adjustable(is_vertical)
+        way_points = grid.get_all_guided_lawnmower_waypoints_adjustable(direction)
```
AC408, line 45: grid.find_border_node() and waypoints = ... can be moved outside of if statements to reduce duplicate code

AC408 (changes requested):
a few comments

tests/functionality_tests/activate_shape_graphs/up_triangle.png
jinlin270: yep

jinlin270: Up and Down for triangle are not working

PR #151, opened 2023-11-04
Separated tests for traversals into files based on activation type

ahanbm:
### Description

I added a file for each of the following 5 activation types: triangle, rectangle, line, grid, square. Previously, the user had to input the traversal type (up, down, left, right) in all capitals. I added logic that makes the check case insensitive (so "up" is matched to "UP") and allows for numeric inputs as well (with "0" corresponding to "RIGHT" and going counterclockwise). I separated the files very simply, just copying the original file (test_activation_shapes) over several times, deleting the irrelevant if conditions and only keeping the code within the relevant if condition. The input logic code is the same in each file and attached below: 
https://drive.google.com/file/d/1evauK4-P5IKvZ0Jx9izkpuCxdHk_Q45W/view?usp=sharing

## Test Plan

The code has been tested by hand for each file with comparison to the original file to make sure everything looks the same. I tested each of the 4 possibilities (up, down, left, right) for each of the 5 activation types. The code works because it was a relatively straightforward refactoring from the original file basically stripping away a layer of complexity.

### DIFF
--- a/engine/grid.py
+++ b/engine/grid.py
@@ -389,6 +389,9 @@ class Grid:
         self.border_nodes = border_list
         self.leftmost_node = leftmost_node
         self.bottom_left_pos = bottom_left_pos
+        self.top_left_pos = top_left_pos
+        self.bottom_right_pos = bottom_right_pos
+        self.top_right_pos = top_right_pos
 
         self.top_left_pos = top_left_pos
         print(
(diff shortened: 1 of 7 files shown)


jinlin270 (changes requested):
The files look good, but please revert the changes to grid.py!

jinlin270 (changes requested):
The tests look good. If we revert the changes to grid.py then we can probably merge it first

ahanbm:
Changing the branch we are merging from to fix dependency issues.

PR #154, opened 2023-11-18
Additional shapes for testing activation and traversal

ahanbm:
### Description

Right now the valid activation types are grid, rectangle, circle, triangle, line. I felt that these types were too specific and the traversal algorithm wouldn't work generally. To verify this I created more corner case type shapes and wrote test code as well to check how the traversal looked. I did not change any of the original code. Traversal worked well on the hexagon, but not on the snake or parallelogram. 

Snake traversal image: https://drive.google.com/file/d/1KwxfJaPlcWRXDclOtR-YANAtaYY4M7fI/view?usp=sharing
Parallelogram traversal image: https://drive.google.com/file/d/1XSxxPk_XBg8EtZba4zWMgAOH3Kf_Lq0x/view?usp=sharing
Hexagon traversal image: https://drive.google.com/file/d/1_X7fpWO-PPDRZhAY6xpZqpe1n27VigEk/view?usp=sharing

## Test Plan

I wrote one "if" statement for each shape I created in test_activation_shapes, consistent with the level of testing of the previous shapes. My methods are implemented so that the shape can be created very generally and additional tests can easily be added. I coped the format of the previous tests to create mine, mainly changing the call to which shape to activate. I visually checked that the right shape was created and analyzed the traversal, which is shown in the pictures above. 
run:
input: parallelogram, snake, hexagon


jinlin270 (changes requested):
Please rebase your PR onto main before merging! This branch is quite behind

ahanbm:
I think I rebased my branch onto main but I'm not sure

Shreya1233 (changes requested):
Great PR! Love the new shapes. Just make sure you document that traversal in general doesn't work for snake or that horizontal traversal doesn't work for parallelogram. We can include those as future tasks for next semester. 

I noticed in the beginning of the vertical traversal for the hexagon, the traversal first traverses on a blue node (inactive waypoint), which it shouldn't do. I think we could resolve this problem by starting the traversal on the next border node instead of the one it starts on? 

Screen Shot 2023-12-02 at 4 41 40 AM

Also a general note: the turns for snake would be very sharp, which would be kind of hard to execute on our robot. Maybe another future todo is adding extra surrounding border nodes when the robot's about to turn in the snake shape so that it can execute the sock-like turn that it does for other shapes? Similarly, I noticed some of the turns were also kind of sharp in the parallelogram as well. I think making it slightly bigger would resolve this problem?

jinlin270 (approved):
Great PR! Rebase, squash commits, and then merge :)

vinnyh125 (approved):
Looks very good! Nice comments detailing what the new functions defined do.

PR #156, opened 2023-11-29
Break beam implementation

Andrew-Lowitt:
Took out all hard-coded attributes in the break-beam sensor and replaced them. Implemented sensor initialization.

### DIFF
--- a/electrical/breakbeam.py
+++ b/electrical/breakbeam.py
@@ -2,27 +2,29 @@ import RPi.GPIO as GPIO
 import time
 from enum import Enum
 
-
 class Breakbeam:
-    ACCEPTABLE_TIME = 3 # time to wait to confirm a full break
-    class BEAM_PINS(Enum):
-        HALF1 = 17
-        HALF2 = 22
-        FULL1 = 23
-        FULL2 = 24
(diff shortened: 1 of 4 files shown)


AC408 (changes requested):
theres a few merge conflicts that needs to be addressed

electrical/breakbeam.py
```diff
@@ -2,27 +2,29 @@
...
-        HALF2 = 22
-        FULL1 = 23
-        FULL2 = 24
+    ACCEPTABLE_TIME = 3  # time to wait to confirm a full break
+
+    # class BEAM_PINS(Enum):
+    #     HALF1 = 17
+    #     HALF2 = 22
+    #     FULL1 = 23
+    #     FULL2 = 24
```
jinlin270, line 13: Why did we remove this class? It looks like it is needed in sensor modules.

jinlin270 (changes requested):

electrical/breakbeam.py
Andrew-Lowitt: This was the hardcoded issue that wouldn't allow me to initialize the sensors elsewhere

PR #158, opened 2023-11-30
Refactor travel

AC408:
## Summary
Refactored travel to also move when not in simulation

### Description
- Updated parameters in robot_state and mission_state
- Adjusted travel to travel based on time period of ms rather than s
- Updated test cases to reflect new travel implementation
- Added movement to travel for when robot not in simulation
- Removed duplicate truthposes

## Test Plan
- `python3 -m engine.main`
- `python3 -m tests.compilation_tests.test_runner`


AC408:
> I love the change! The code is so much cleaner by moving the check for is_sim inside of the travel funciton. Also I never knew we had this cool simulation??? You gotta put it on the CDR

wait, which simulation? the one from calling `python3 -m main.engine`?

Shreya1233 (approved):
Great PR!

PR #159, opened 2023-12-01
Refactor Grid Traversal to use a simpler algorithm

jinlin270:
## Summary
Switch to use a very simple algorithm for lawnmower traversal by keeping track of the activated nodes in a 2D array during activation shape.

### Description
The algorithm works as follows: Suppose we have a grid like the one below, and we want to traverse the red activated region. During "activation", we will store the activated nodes into a 2D array by rows. If we want to traverse, we simply add one row at a time to the waypoints, and we turn and traverse in the opposite direction by reversing the order of the points in every other row. 
To traverse RIGHT, we flip every even row.
To traverse LEFT, we flip every odd row.
Similar logic applies for vertical traversal, by keeping track of a 2D array sorted by columns during activation.

IMG_3E96D7D0A3FA-1

Other changes:
* The algorithm keeps track of each node as a Node object, rather than a tuple of coordinates. This change is necessary during real life traversal so the longitude and latitude can be accessed.
* In practice, the algorithm only add the endpoints of each row or column, removing the unnecessary points in the middle to be more pid friendly.

Pros:
This algorithm gives a more intuitive understanding of the lawnmower traversal, and dramatically decreases the amount of code needed in grid.py. 
Cons:
The drawback to the reduced complexity is the increase in needed space by storing the 2D arrays of activated nodes. However, it shouldn't be too much of an issue considering we need to keep track of the entire grid  of nodes regardless, so the space complexity remains the same. It can also be mitigated if we choose to store only the endpoints of each row or column during activation
## Test Plan

Activation type: circle, rectangle, triangle, line
Traversal Direction: LEFT, RIGHT, DOWN, UP

## Other

TODO:
1. Implement turning algorithm to yield a list of Nodes. 
2. Retire prior code in grid.py if we do move forward with this algorithm
3. If want to further optimize, change the activation shapes so instead of storing the entire row or columns, store only the end points

Credits to Ahan for the code in test_activation_shapes.py!

### DIFF
--- a/engine/grid.py
+++ b/engine/grid.py
@@ -44,7 +44,7 @@ class Grid:
 
         #num_rows: Number of rows of Nodes in the grid [int]
         #num_cols: Number of columns of Nodes in the grid. [int]
-
+        #TODO: remove leftmost_node, leftmost_node_pos, border_nodes, 
         #leftmost_node: the leftmost active node in the Grid which is used as the starting node in lawnmower and spiral traversal.
         #leftmost_node_pos: the (row,col) position of the leftmost node
         #border_nodes: all active nodes which either exist on the edge of the grid or have a neighbor that is an inactive node
@@ -150,6 +150,9 @@ class Grid:
         self.curr_pos = None
         self.direction = self.Direction.RIGHT
         self.waypoints = []
+        # RFWP
+        self.vertical_waypoint_list = []
+        self.horizontal_waypoint_list = [[] for _ in range(self.num_cols)]
 
     def get_active_waypoints_list(self):
         return self.active_waypoints_list
@@ -219,9 +222,15 @@ class Grid:
         Activates all the nodes in a rectangle based on row start/end, col
         start/end.
         """
+        self.horizontal_waypoint_list = [[] for _ in range(col_limit - col)]
         for x in range(row, row_limit):
+            # RFWP
+            col_list = []
             for y in range(col, col_limit):
                 self.nodes[x, y].is_active = True
+                col_list.append(self.nodes[x, y])
+                self.horizontal_waypoint_list[y - col].append(self.nodes[x, y])
+            self.vertical_waypoint_list.append(col_list)
         return self.nodes
 
     def activate_circle(self, circle_center_row, circle_center_col, circle_radius):
@@ -229,6 +238,7 @@ class Grid:
         Activates all the nodes in a circle based on center and radius.
         """
         for x in range(self.num_rows):
+            col_list = []
             for y in range(self.num_cols):
                 if (
                     (x - circle_center_row) ** 2
@@ -236,6 +246,10 @@ class Grid:
                     - circle_radius**2
                 ) < 0:
                     self.nodes[x, y].is_active = True
+                    col_list.append(self.nodes[x, y])
+                    self.horizontal_waypoint_list[y].append(self.nodes[x, y])
+            if col_list:
+                self.vertical_waypoint_list.append(col_list)
 
     def is_inside_triangle(self, p1, p2, p3, p4):
         """A function to check whether point p4 lies inside the triangle formed by
@@ -267,31 +281,38 @@ class Grid:
         # is same as A
         return A == (A1 + A2 + A3)
 
-    def activate_line(self, coordinate, n, is_horizontal):
+    def activate_line(self, coordinate, n: int, is_vertical: bool):
         """
         Activate a single horizontal line of length n starting at row, col.
         If is_horizontal is True, the line is from (row, col) to (row, col + n).
         Otherwise, the line is from (row, col) to (row + n, col).
         Args:
             coordinate (int tuple): coordinate of the starting node
+            n: the number of points in the line
+            is_vertical: whether the line is vertical or horizontal
         """
         (row, col) = coordinate
         # Note: rows are y-position and columns are x-position
-        if is_horizontal:
-            for i in range(n):
+        for i in range(n):
+            if is_vertical:
+                self.nodes[col, row + i].is_active = True
+                self.vertical_waypoint_list.append([self.nodes[col, row + i]])
+            else:
                 self.nodes[col + i][row].is_active = True
-        else:
-            for i in range(n):
-                self.nodes[col][row + i].is_active = True
+                self.horizontal_waypoint_list.append([self.nodes[col + i, row]])
 
     def activate_triangle(self, p1, p2, p3):
         """
-        Activates all the nodes in a triangle based on three vertices p1, p2, p3.
+        Activates all the nodes in a triangle based on three vertices p1, p2, p3. The vertices should all be on the positive x-y quadrant
         """
         for x in range(self.num_rows):
+            col = []
             for y in range(self.num_cols):
                 if self.is_inside_triangle(p1, p2, p3, (x, y)):
                     self.nodes[x, y].is_active = True
+                    self.horizontal_waypoint_list[y].append(self.nodes[x, y])
+                    col.append(self.nodes[x, y])
+            self.vertical_waypoint_list.append(col)
 
     def activate_parallelogram(self, row, col, row_limit):
         # Horizontal traversal does not currently work for this shape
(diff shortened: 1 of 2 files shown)


engine/grid.py
```diff
@@ -248,31 +282,38 @@ def area(p1, p2, p3):
         # is same as A
         return A == (A1 + A2 + A3)
 
-    def activate_line(self, coordinate, n, is_horizontal):
+    def activate_line(self, coordinate, n, is_vertical):
         """
         Activate a single horizontal line of length n starting at row, col.
```
AC408, line 287: update specs

engine/grid.py
```diff
@@ -248,31 +282,38 @@ def area(p1, p2, p3):
...
+        if not is_vertical:
             for i in range(n):
-                self.nodes[col+i][row].is_active = True
+                self.nodes[col + i][row].is_active = True
+                self.horizontal_waypoint_list.append([self.nodes[col + i, row]])
         else:
             for i in range(n):
-                self.nodes[col][row+i].is_active = True
+                self.nodes[col][row + i].is_active = True
+                self.vertical_waypoint_list.append([self.nodes[col, row + i]])
```
AC408, line 302: for loop, then conditional expression for faster computation

engine/grid.py
```diff
@@ -248,31 +282,38 @@ def area(p1, p2, p3):
...
 
     def activate_triangle(self, p1, p2, p3):
         """
         Activates all the nodes in a triangle based on three vertices p1, p2, p3.
         """
+        self.horizontal_waypoint_list = [[] for _ in range(self.num_cols)]
```
AC408, line 308: can we initialize this in the class constructor and only need to change it for activate_rectangle?

engine/grid.py
```diff
@@ -211,26 +222,49 @@ def activate_rectangle(self, row, col, row_limit, col_limit):
         Activates all the nodes in a rectangle based on row start/end, col
         start/end.
         """
+        self.horizontal_waypoint_list = [[] for _ in range(col, col_limit)]
```
AC408, line 225: is col_limit - col = num_col?

engine/grid.py
```diff
@@ -211,26 +222,49 @@ def activate_rectangle(self, row, col, row_limit, col_limit):
...
+            self.vertical_waypoint_list.append(col_list)
         return self.nodes
 
     def activate_circle(self, circle_center_row, circle_center_col, circle_radius):
         """
         Activates all the nodes in a circle based on center and radius.
         """
+        self.horizontal_waypoint_list = [[] for _ in range(self.num_cols)]
         for x in range(self.num_rows):
+            col = []
```
AC408, line 242: nit, naming invonsistencies

engine/grid.py
```diff
@@ -600,6 +699,58 @@ def add_guided_waypoints(self, is_vertical):
...
+                if flip:
+                    self.waypoints += [
+                        self.vertical_waypoint_list[row][-1],
+                        self.vertical_waypoint_list[row][0],
+                    ]
+                else:
+                    self.waypoints += [
+                        self.vertical_waypoint_list[row][0],
+                        self.vertical_waypoint_list[row][-1],
+                    ]
```
AC408, line 728: funny way to simplify:
```
tmp = self.vertical_waypoint_list[row] 
self.waypoints += [tmp[-flip], tmp[-(not flip)]]
```
not sure if its a good change

engine/grid.py
```diff
@@ -600,6 +699,58 @@ def add_guided_waypoints(self, is_vertical):
...
+                if flip:
+                    self.waypoints += [
+                        self.horizontal_waypoint_list[col][-1],
+                        self.horizontal_waypoint_list[col][0],
+                    ]
+                else:
+                    self.waypoints += [
+                        self.horizontal_waypoint_list[col][0],
+                        self.horizontal_waypoint_list[col][-1],
+                    ]
```
AC408, line 750: can be made into helper function

engine/grid.py
```diff
@@ -600,6 +699,58 @@ def add_guided_waypoints(self, is_vertical):
...
+                flip = not flip
```
AC408, line 751: can be put outside if else

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -18,13 +17,26 @@ def main():
         print("flip rows, cols for traversal")
 
     activation_type = input("Activation type: ")
-    is_vertical = "True" == input("Vertical traversal: ")
-    if is_vertical:
+    direction = input("Traversal Direction: ")
+    if direction.isnumeric():
+        numberToDirection = {"0": "RIGHT", "1": "UP", "2": "LEFT", "3": "DOWN"}
```
AC408, line 22: im ngl, i dont know if this is necessary to add; i dont know who would associate 0 to right, 1 to up, etc

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -18,13 +17,26 @@ def main():
...
+        direction = numberToDirection.get(direction)
+    else:
+        direction = direction.upper()
+
+    print(direction)
```
AC408, line 27: remove print

tests/functionality_tests/test_activation_shapes.py
```diff
@@ -18,13 +17,26 @@ def main():
...
+    if direction == "UP":
         grid.direction = grid.Direction.UP
+    elif direction == "DOWN":
+        grid.direction = grid.Direction.DOWN
+    elif direction == "RIGHT":
+        grid.direction = grid.Direction.RIGHT
+    else:
+        grid.direction = grid.Direction.LEFT
```
AC408, line 35: i think else condition should be elif and add raise error for else

AC408 (changes requested):
minor comments; can you resolve merge conflicts before making pr?

engine/grid.py
jinlin270: This is also needed for shapes like circle or any slanted shapes we may have

engine/grid.py
jinlin270: I think? I was following the logic in the original code

engine/grid.py
jinlin270: Done

tests/functionality_tests/test_activation_shapes.py
jinlin270: Done

tests/functionality_tests/test_activation_shapes.py
jinlin270: It doesn't hurt to keep it? Also credits to Ahan! I test_activation_shapes.py from his PR and it's great.

engine/grid.py
jinlin270: Smart! not as readable though... I think we can afford to keep it without the optimization

engine/grid.py
AC408: yeah, but activate_circle uses that exact same initialization

tests/functionality_tests/test_activation_shapes.py
AC408: ig not, i dont know if it'll crowd the code space

engine/grid.py
jinlin270: Good point

engine/grid.py
```diff
@@ -248,31 +282,38 @@ def area(p1, p2, p3):
...
         """
+        self.horizontal_waypoint_list = [[] for _ in range(self.num_cols)]
         for x in range(self.num_rows):
+            col = []
             for y in range(self.num_cols):
                 if self.is_inside_triangle(p1, p2, p3, (x, y)):
                     self.nodes[x, y].is_active = True
+                    self.horizontal_waypoint_list[y].append(self.nodes[x, y])
+                    col.append(self.nodes[x, y])
+            self.vertical_waypoint_list.append(col)
```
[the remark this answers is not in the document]
jinlin270: There doesn't seem like this is an issue as long as is_inside_triangle is correct (which I believe it is with the condition that the points are on the positive x,y coordinate. I added the pre-condition for now, and we can make it a future todo to make it work for the entire x-y plane if needed. 
I am not sure if this is what you are talking about?

engine/grid.py
jinlin270: thank you! Done

ahanbm (changes requested):
Couple small changes: Can you remove the comments above simple_get_guided_waypoints (or make it into a separate method if you think the previous method might still have some utility). Can you also add a short bit of documentation to the simple_get_guided_waypoints method.

AC408 (approved):

ahanbm (approved):
Looks good!

